PR #1, opened 2020-03-24
Bring code from element-rtp

bblaszkow06:
The code here is based on membrane-element-rtp from master with rtcp cherry-picked from rtcp (to skip srtp). Therefore, it's mostly still WIP, but to make future PRs smaller I'd like a review now. The main focus should be on module naming and directory organization


lib/membrane/rtcp/app.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Membrane.RTCP.App do
```
mat-hek: It's a bit ambiguous, we call application modules like that. I would make it `AppPacket`

lib/membrane/rtcp/bye.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Membrane.RTCP.Bye do
```
mat-hek: ByePacket

lib/membrane/rtcp/packet.ex
```diff
@@ -0,0 +1,100 @@
+defmodule Membrane.RTCP do
```
mat-hek: seems duplicate with packet_parser.ex

lib/membrane/rtcp/report_block.ex
```diff
@@ -0,0 +1,65 @@
+defmodule Membrane.RTCP.ReportBlock do
```
mat-hek: if it belongs to Report, should be Report.Block

lib/membrane/rtp/header_extension.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Membrane.RTP.HeaderExtension do
```
mat-hek: Header.Extension ?

lib/membrane/rtcp/packet_parser.ex
```diff
@@ -0,0 +1,100 @@
+defmodule Membrane.RTCP.PacketParser do
```
mat-hek: Would call it just Packet

lib/membrane/rtp/header.ex
```diff
@@ -0,0 +1,72 @@
+defmodule Membrane.RTP.Header do
```
mat-hek: Packet.Header

lib/membrane/rtp/packet_parser.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Membrane.RTP.PacketParser do
```
mat-hek: merge to RTP.Packet

lib/membrane/rtp/parser_element.ex
```diff
@@ -0,0 +1,97 @@
+defmodule Membrane.RTP.Parser do
```
mat-hek: adjust file name

lib/membrane/rtcp/sdes.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Membrane.RTCP.SDES do
```
bblaszkow06: SDESPacket to be consistent.

lib/membrane/rtcp/packet_parser.ex
bblaszkow06: OK, that would solve naming conflict with parser element :+1:

lib/membrane/rtcp/app.ex
mat-hek: What's behind this `Specific` suffix? IMHO, the best way would be to append `Packet` to everything but for Report, but even having ReportPacket is clearer than having `.App` or `.Bye`

mat-hek: Or maybe RTCP.Packet**s**.App

bblaszkow06: Specific comes from the RFC: `APP:  Application-specific functions`. It could also be `Defined` suffix: `6.7 APP: Application-Defined RTCP Packet` Anyway, I'll add Packet suffix, after all, these modules won't be used often outside the parser

bblaszkow06: + lone Packet might be a bit confusing

lib/membrane/rtp/header_extension.ex
mat-hek: still unchanged

lib/membrane/rtp/payload_type_decoder.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Membrane.RTP.PayloadTypeDecoder do
```
mat-hek, line 1: It should be a part/submodule of RTP.Packet. Or at least it shouldn't end with Decoder

lib/membrane/rtcp/packet.ex
```diff
@@ -1,100 +1,41 @@
...
+defmodule Membrane.RTCP.Packet do
   @moduledoc """
-  Parses compound RTCP and SRTCP packets into a list of subpackets.
+  Functions common to all RTCP Packets
   """
 
-  # parse_compound - authenticates/decryptes an entire compound packet
-  # parse - removes padding from compound packet
-  # parse_subpacket - parses an extracted binary into a single subpacket (single SR/RR/BYE/etc)
+  # TODO: Add a behaviour for packets
```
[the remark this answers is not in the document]
bblaszkow06: It is as it now exists but only implicitly (https://github.com/membraneframework/membrane_rtp/blob/0ac63503b1028de315a966d720d8d4795664e0b1/lib/membrane/rtcp/compound_packet.ex#L84-L89). It's not too much work and might help in future refactors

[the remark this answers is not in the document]
mat-hek: imho adding such behaviour would not help improve this code in any way, while it would introduce some boilerplate

PR #3, opened 2020-04-07
JitterBuffer & Bin

bblaszkow06:
Brings code from jitterbuffer and bin-rtp
Uses the renamed RTP caps as well as depayloaders.


test/support/jitter_buffer/buffer_factory.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Membrane.RTP.JitterBufferTest.BufferFactory do
```
mat-hek, line 1: factory, tfu!

PR #8, opened 2020-07-14
Rtcp sending

bblaszkow06:
- [x] send reports basing on jitterbuffer stats
- [x] add inbound RTCP data to reports
- [x] docs


lib/membrane/rtcp/receiver_reporter.ex
```diff
@@ -0,0 +1,154 @@
...
+                """
+              ],
+              sender_report_timeout: [
+                type: :time,
+                default: 60 |> Time.seconds(),
+                description: """
+                Minimal time between receiving RTCP sender report
+                and assuming it won't be needed anymore.
+                """
+              ]
```
bblaszkow06, line 27: This seems like a dirty fix rather than proper solution. The parent should inform about ssrc pad removal.

In general, I have the feeling that half of this element should not be an element but rather a module used by receiver bin. The responsibility of this element could just push the RTCP reports and Receiver Bin, instead of forwarding numerous messages to the Reporter would simply call functions.

Imagine that we wanted a proper report scheduling based on bandwidth - that would be yet another set of messages (some kind of bandwidth reports) to forward here in order to calculate when the report should be sent

lib/membrane/rtp/jitter_buffer.ex
```diff
@@ -181,16 +191,13 @@ defmodule Membrane.RTP.JitterBuffer do
...
-    use Bitwise
+  @spec get_updated_stats(State.t()) :: {Stats.t(), State.t()}
+  defp get_updated_stats(state) do
+    %State{store: store, stats_acc: stats_acc} = state
 
     # Variable names follow algorithm A.3 from RFC3550 (https://tools.ietf.org/html/rfc3550#appendix-A.3)
-    %{base_index: base_seq, end_index: extended_max, received: received} = store
-    expected = extended_max - base_seq + 1
+    %BufferStore{base_index: base_seq, end_index: extended_max, received: received} = store
+    expected = if base_seq == nil, do: 0, else: extended_max - base_seq + 1
```
bblaszkow06, line 200: This seems like trouble - what kind of stats we can generate if nothing was received (only then base_seq is `nil`)? Instead of doing this if, I'd rather prevent calling this function in such state

test/membrane/rtp/session/receive_bin_test.exs
```diff
@@ -13,17 +13,60 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
 
   @fmt_mapping %{96 => {:H264, 90_000}, 127 => {:MPA, 90_000}}
 
+  defmodule Pauser do
+    # move to Core.Testing?
```
bblaszkow06, line 17: Start by writing some moduledoc

lib/membrane/rtp/jitter_buffer.ex
mat-hek: > what kind of stats we can generate if nothing was received

Empty stats. We have to reply somehow for the stats request, why not reply with empty stats?

lib/membrane/rtcp/receiver_reporter.ex
```diff
@@ -0,0 +1,154 @@
...
+    if Enum.empty?(state.ssrcs) do
+      buffer = %Buffer{payload: make_report(state)}
+      {{:ok, buffer: {:output, buffer}}, %{state | stats: []}}
+    else
+      {:ok, state}
+    end
+  end
+
+  @impl true
+  def handle_other({:rtcp, rtcp}, _ctx, state) do
```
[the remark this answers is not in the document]
mat-hek: good point, make an issue and propose a name :)

test/membrane/rtp/session/receive_bin_test.exs
mat-hek: done

lib/membrane/rtp/jitter_buffer.ex
bblaszkow06: Struct filled with 0s is not a good indicator of "not ready to send stats, don't include it in RTCP packet"

PR #11, opened 2020-07-29
Credo linting

bblaszkow06:


Qizot:
Credo has already been added as this draft got pretty old.

PR #17, opened 2020-08-14
Rtcp reporting refactor

mat-hek:


lib/membrane/rtp/session/report.ex
```diff
@@ -0,0 +1,137 @@
...
+    maybe_report =
+      if Enum.empty?(report_data.remote_ssrcs) do
+        :no_report
+      else
+        Membrane.Logger.warn(
+          "Not received stats from ssrcs: #{Enum.join(report_data.ssrcs, ", ")}"
+        )
+
+        {:report, generate_report(report_data)}
+      end
```
bblaszkow06, line 41: Sending the report here doesn't seem like a good idea
- timeout for stats gathering should be separate from frequency of report generation
- if after this report the stats will be gathered quickly, we'd generate 2 reports one after another
- sending report from `init_report` is not quite intuitive

I would also consider changing the way reports are obtained: returned value would just contain atom indicating whether the report is ready and a separate function for obtaining it that would be used both when message is ready and when the timeout is reached

lib/membrane/rtp/session/receive_bin.ex
```diff
@@ -188,11 +190,28 @@ defmodule Membrane.RTP.Session.ReceiveBin do
...
+
+    stats_requests = Enum.map(remote_ssrcs, &{{:rtp_stream_bin, &1}, :send_stats})
+
+    report_messages =
+      case maybe_report do
+        {:report, report} -> [rtcp_forwarder: {:report, report}]
+        :no_report -> []
+      end
+
+    actions = Enum.map(report_messages ++ stats_requests, &{:forward, &1})
```
bblaszkow06, line 209: Forward action will not accept list? If not, this should be added to the issue in core regarding a rewrite of this action

lib/membrane/rtp/session/report.ex
```diff
@@ -0,0 +1,137 @@
...
+  def handle_stats(stats, remote_ssrc, ssrcs, report_data) do
+    report_ssrcs = MapSet.delete(report_data.remote_ssrcs, remote_ssrc)
+
+    stats =
+      case Map.fetch(ssrcs, remote_ssrc) do
+        {:ok, local_ssrc} -> [{local_ssrc, remote_ssrc, stats}]
+        :error -> []
+      end
+
+    report_data = %{report_data | stats: stats ++ report_data.stats, remote_ssrcs: report_ssrcs}
```
bblaszkow06, line 74: Lists concatenation when building them should be avoided. Does the order of stats matter? If you still want to use `++` you can at least swap arguments, so that LHS will always have 1 element at most.

lib/membrane/rtp/session/receive_bin.ex
```diff
@@ -217,41 +236,31 @@ defmodule Membrane.RTP.Session.ReceiveBin do
...
+        ctx,
+        state
+      ) do
+    {result, report_data} =
+      Session.Report.handle_stats(stats, remote_ssrc, state.ssrcs, state.rtcp_report_data)
+
+    state = %{state | rtcp_report_data: report_data}
+
+    case {result, ctx.children} do
+      {{:report, report}, %{rtcp_forwarder: _reporter}} ->
```
bblaszkow06, line 258: This match isn't pretty. It would be much better if you could replace it with something more self descriptive like:
`if result == :ready and ctx.children |> Map.has_key(:rtcp_forwarder)`

lib/membrane/rtp/session/report.ex
mat-hek: > timeout for stats gathering should be separate from frequency of report generation

I believe that would only introduce unnecessary complexity. Stats should be always gathered quickly, and if they don't come till the next tick, there's a problem with the frequency, not the stats.

> if after this report the stats will be gathered quickly, we'd generate 2 reports one after another

that shouldn't happen, and even if it happens to, it doesn't hurt us

> I would also consider changing the way reports are obtained: returned value would just contain atom indicating whether the report is ready and a separate function for obtaining it that would be used both when message is ready and when the timeout is reached

Then we should support the case when one doesn't generate a report when it's ready. And I don't see any benefits of that.

lib/membrane/rtp/session/receive_bin.ex
mat-hek: You mean forward should support forwarding the same message to multiple children or a list of {child, message} tuples or both?

lib/membrane/rtp/session/report.ex
mat-hek: well, LHS has one element at most...

lib/membrane/rtp/session/receive_bin.ex
bblaszkow06: > a list of {child, message} tuples

lib/membrane/rtp/session/report.ex
bblaszkow06: > introduce unnecessary complexity

The only unnecessary complexity I see is doing some weird stuff in `init_report` :man_shrugging: 

> Stats should be always gathered quickly, and if they don't come till the next tick, there's a problem with the frequency, not the stats

The keyword is "SHOULD". The assumption that it has to be a frequency problem is wrong. What if you have a bug causing JitterBuffer to get stuck? Or the message will not reach the target because of bug in some bin on the way? Or the bin's message queue is huge? Or the system is overloaded? Or maybe the SSRC was removed after sending a stats request? Besides, at some point we have to combine report sending frequency with the bandwidth and mixing that logic with stats gathering timeout is more of the unnecessary complexity.

>  it doesn't hurt us

It does break the bandwidth constraints from RFC. Of course, it's not a critical failure, however this is still a bug, shouldn't work this way and I don't see any benefit of not fixing it.

> when one doesn't generate a report when it's ready

What's here to support? Init should clear the state anyway and such scenario would occur if the bandwidth constraints would disallow sending the report.

lib/membrane/rtp/session/receive_bin.ex
```diff
@@ -183,7 +183,6 @@ defmodule Membrane.RTP.Session.ReceiveBin do
 
   @impl true
   def handle_pad_removed(Pad.ref(:output, ssrc), _ctx, state) do
-    # TODO: parent may not know when to unlink, we need to timout SSRCs and notify about that and BYE packets over RTCP
```
bblaszkow06, line 184: Was that really solved?

lib/membrane/rtp/session/report.ex
mat-hek: > The only unnecessary complexity I see is doing some weird stuff in init_report 🤷‍♂️

The other is adding useless timeouts and timers

> The keyword is "SHOULD" (...)

If we encounter any of these bugs, we will really have bigger problems than exceeding RTCP bandwidth constraints

Nevertheless, If you think we need an additional timeout for stats gathering, how should it be obtained/calculated?

lib/membrane/rtp/session/report.ex
```diff
@@ -48,11 +38,22 @@ defmodule Membrane.RTP.Session.Report do
...
-        stats: []
+        remote_reports: remote_reports
     }
 
-    {maybe_report, remote_ssrcs, report_data}
+    {remote_ssrcs, report_data}
+  end
+
+  @spec ensure_nothing_to_report(Data.t()) :: {maybe_report_t, Data.t()}
+  def ensure_nothing_to_report(report_data) do
```
bblaszkow06, line 48: Still no timeout, but I can live with that. However, I'd change the name to `flush_report`

bblaszkow06 (approved):

PR #22, opened 2020-11-20
Sending sender reports

### LINKED ISSUE #21: Generate sender reports
according to https://tools.ietf.org/html/rfc3550#section-6.4

AndrzejPodo:
What inside:
* add collecting and sending SR(Sender Reports)
* attempts to solve problem with RTCP report receiving test in `session_bin_test.ex` by adding `RTCPPacketInspector` filter which sends notification when all reports of specified pattern were seen at least once.

Problems:
* problem with RTCP receiver reports that was accidentally found is not fully investigated and solved in that PR. 
* timestamps in SR are not set properly yet.

closes #21


test/membrane/rtp/session_bin_test.exs
```diff
@@ -27,6 +31,120 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
   }
 
   @fmt_mapping %{96 => {:H264, 90_000}, 127 => {:MPA, 90_000}}
+  defmodule RTCPPacketInspector do
```
mat-hek, line 34: let's move that to the test itself

test/membrane/rtp/session_bin_test.exs
```diff
@@ -27,6 +31,120 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
...
+            actions ++ [{:notify, :all_sr_received}]
+
+          _ ->
+            actions
+        end
+
+      {{:ok, actions}, state}
+    end
+
+    defp compare_on_common_fields(left, right) when is_struct(left) when is_struct(right) do
```
mat-hek, line 118: This clause makes this function not really comparing on common fields since it ignores the `__struct__` field. Since you seem not to use that anyway, maybe remove this clause?

test/membrane/rtp/session_bin_test.exs
```diff
@@ -27,6 +31,120 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
...
+          true -> Map.from_struct(left)
+          _ -> left
+        end
+
+      compare_on_common_fields(left, right)
+    end
+
+    defp compare_on_common_fields(left, right) when is_map(left) and is_map(right) do
+      left
+      |> Enum.reduce(true, fn {k, v}, acc ->
```
mat-hek, line 136: use `Enum.all`

mat-hek (changes requested):

lib/membrane/rtp/session/sender_report.ex
```diff
@@ -0,0 +1,109 @@
+defmodule Membrane.RTP.Session.SenderReport do
```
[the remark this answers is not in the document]
AndrzejPodo: yes, sure, was also thinking about that 👍

test/membrane/rtp/session_bin_test.exs
AndrzejPodo: Ok, the reason this clause was introduced because is that it's not allowed to iterate over `{k,v}` pairs in struct, so it was only ensuring that both arguments are maps(which implements Enumerable). Although I agree the action that function takes is not consistent with the naming. Anyway I'm going to completely rewrite it. Thanks.

lib/membrane/rtp/serializer.ex
```diff
@@ -65,6 +92,38 @@ defmodule Membrane.RTP.Serializer do
...
+  end
+
+  @spec get_updated_stats(State.t()) :: {Stats.t(), State.t()}
+  defp get_updated_stats(%State{any_buffer_sent?: false} = state) do
+    {:no_stats, state}
+  end
+
+  defp get_updated_stats(%State{stats_acc: stats} = state) do
+    {stats, state}
+  end
```
mat-hek, line 119: This actually doesn't update stats, `get_stats` would be more accurate. Also, I don't see any reason to return state if it never changes

lib/membrane/rtp/serializer.ex
```diff
@@ -65,6 +92,38 @@ defmodule Membrane.RTP.Serializer do
...
+
+  defp update_counters(%Buffer{payload: payload}, state) do
+    state
+    |> put_in(
+      [:stats_acc, :sender_octet_count],
+      state.stats_acc.sender_octet_count + Payload.size(payload)
+    )
+    |> put_in([:stats_acc, :sender_packet_count], state.stats_acc.sender_packet_count + 1)
```
mat-hek, line 127: update_in would fit better

lib/membrane/rtp/session/sender_report.ex
```diff
@@ -0,0 +1,109 @@
...
+  def handle_stats(stats, sender_ssrc, data) do
+    senders_ssrcs = MapSet.delete(data.senders_ssrcs, sender_ssrc)
+
+    data = %{data | stats: Map.put(data.stats, sender_ssrc, stats), senders_ssrcs: senders_ssrcs}
+
+    if Enum.empty?(senders_ssrcs) do
+      sender_reports = generate_report(data.stats)
+
+      case sender_reports.packets do
+        [] -> {:no_report, data}
```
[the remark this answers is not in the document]
mat-hek: bump

lib/membrane/rtp/session_bin.ex
```diff
@@ -383,18 +387,41 @@ defmodule Membrane.RTP.SessionBin do
 
   @impl true
   def handle_tick(:rtcp_report_timer, _ctx, state) do
-    {maybe_report, report_data} = Session.Report.flush_report(state.rtcp_report_data)
-    {remote_ssrcs, report_data} = Session.Report.init_report(state.ssrcs, report_data)
-    stats_requests = Enum.map(remote_ssrcs, &{{:stream_receive_bin, &1}, :send_stats})
+    {maybe_report, report_data} = Session.ReceiverReport.flush_report(state.rtcp_report_data)
```
mat-hek, line 390: maybe_receiver_report

lib/membrane/rtp/session_bin.ex
```diff
@@ -383,18 +387,41 @@ defmodule Membrane.RTP.SessionBin do
...
+    {remote_ssrcs, report_data} = Session.ReceiverReport.init_report(state.ssrcs, report_data)
 
-    report_messages =
+    {maybe_senders_report, senders_report_data} =
```
mat-hek, line 393: maybe_sender_report, sender_report_data

test/membrane/rtp/session_bin_test.exs
```diff
@@ -28,6 +32,46 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
...
+    end
+  end
+
+  defp compare(left, right), do: left == right
+
+  # asserts all specified buffers were received by the sink
+  defp assert_specified_buffers(_pipeline, _sink, []), do: :ok
+
+  defp assert_specified_buffers(pipeline, sink, field_specyfications) do
+    assert_sink_buffer(pipeline, sink, buffer, 10000)
```
mat-hek, line 62: why such a big timeout?

test/membrane/rtp/session_bin_test.exs
```diff
@@ -28,6 +32,46 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
...
+    %Membrane.Buffer{payload: <<head::binary-size(4), body_and_rest::binary>>} = buffer
+    {:ok, %{header: header, length: len}} = Header.parse(head)
+    body_size = len - 4
+    <<body::binary-size(body_size), rest::binary>> = body_and_rest
+    {:ok, packet} = Packet.parse_body(body, header)
```
mat-hek, line 67: this is what RTCP.CompoundPacket.parse does

test/membrane/rtp/session_bin_test.exs
```diff
@@ -28,6 +32,46 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
...
+          _ ->
+            false
+        end
+      end)
+    else
+      false
+    end
+  end
+
+  defp compare(left, right), do: left == right
```
mat-hek, line 56:
```suggestion
  @spec compare_common_fields(any(), any()) :: boolean()
  defp compare_common_fields(left, right) when is_map(left) do
    is_map(right) and Enum.all?(right, fn {k, v} ->
        Map.has_key?(left, k) and compare_common_fields(Map.get(left, k), v)
    end)
  end

  defp compare_common_fields(left, right), do: left == right
```

test/membrane/rtp/session_bin_test.exs
```diff
@@ -28,6 +32,46 @@ defmodule Membrane.RTP.Session.ReceiveBinTest do
 
   @fmt_mapping %{96 => {:H264, 90_000}, 127 => {:MPA, 90_000}}
 
+  @spec inspect_packet_fields(Packet.t(), %{}) :: boolean()
+  defp inspect_packet_fields(packet, fields_values), do: compare(packet, fields_values)
```
mat-hek, line 36: Inspect doesn't feel right here, maybe 'match_packet' ?

lib/membrane/rtp/serializer.ex
```diff
@@ -24,13 +25,37 @@ defmodule Membrane.RTP.Serializer do
...
               ]
 
+  defmodule State do
+    @moduledoc false
+    use Bunch.Access
+
+    defstruct sequence_number: 0,
+              init_timestamp: 0,
+              any_buffer_sent?: false,
+              stats_acc: %Stats{
```
mat-hek, line 35: This is actually not `stats_acc` field, but actual `Stats`. Maybe let's call it 'stats' and init with `:no_stats`?

mat-hek (changes requested):

lib/membrane/rtp/serializer.ex
AndrzejPodo: okay

test/membrane/rtp/session_bin_test.exs
AndrzejPodo: Nice, that's much more compact :D But here in line 39 `defp compare_common_fields(left, right) when is_map(left) do` there should be also `or is_struct(left)` as the left side is any structure(map or struct) and right is map of fields we want to match. Do you agree or getting rid of `or is_struct(left)` was intended?

test/membrane/rtp/session_bin_test.exs
AndrzejPodo: Yeah, it can be removed from here, but anyway there is still a problem that sometimes we have to wait quite a few seconds until all RR are collected (as I observed usually SR-s are delivered almost immediately).

lib/membrane/rtp/session_bin.ex
AndrzejPodo: ok

lib/membrane/rtp/session_bin.ex
AndrzejPodo: ok

lib/membrane/rtp/serializer.ex
AndrzejPodo: done

lib/membrane/rtp/serializer.ex
AndrzejPodo: yes, thats right, done.

lib/membrane/rtp/session/sender_report.ex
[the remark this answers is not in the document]
AndrzejPodo: oki, done

test/membrane/rtp/session_bin_test.exs
AndrzejPodo: done

mat-hek (approved):

PR #27, opened 2020-12-03
fix RTP extension header parsing and serialization

### LINKED ISSUE #23: Support RTP extensions properly
It seems that we don't handle RTP extensions properly. If they are set (in SDP offer/answer) our depayloaders receive improper payloads. This happens both for OPUS and H264 depayloaders. Removing RTP extensions from SDP answer seems to resolve problem. 

Example extension set that causes errors in OPUS (audio):

```
    a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
    a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
    a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
    a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
    a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
    a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
```

Example extension set that causes errors in H264 (video):

```
    a=extmap:14 urn:ietf:params:rtp-hdrext:toffset
    a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
    a=extmap:13 urn:3gpp:video-orientation
    a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
    a=extmap:12 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
    a=extmap:11 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
    a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
    a=extmap:8 http://www.webrtc.org/experiments/rtp-hdrext/color-space
    a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:mid
    a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
    a=extmap:6 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
```

The task is to investigate where is the problem.

mat-hek:
closes #23

### DIFF
--- a/lib/membrane/rtp/packet.ex
+++ b/lib/membrane/rtp/packet.ex
@@ -43,7 +43,13 @@ defmodule Membrane.RTP.Packet do
   end
 
   defp serialize_header_extension(%Header.Extension{} = extension) do
-    <<extension.profile_specific::16, byte_size(extension.data)::16, extension.data::binary>>
+    data_size = byte_size(extension.data)
+    padded_data_size = Bunch.Math.min_multiple_gte(4, data_size)
+    padding = padded_data_size - data_size
+    length = div(padded_data_size, 4)
+
+    <<extension.profile_specific::16, length::16, extension.data::binary,
+      0::size(padding)-unit(8)>>
   end
 
   @spec parse(binary()) :: {:ok, t()} | {:error, :wrong_version | :malformed_packet}
@@ -80,7 +86,8 @@ defmodule Membrane.RTP.Packet do
   defp parse_header_extension(binary, false), do: {:ok, {nil, binary}}
 
   defp parse_header_extension(
-         <<profile_specific::16, data_len::16, data::binary-size(data_len), rest::binary>>,
+         <<profile_specific::16, data_len::16, data::binary-size(data_len)-unit(32),
+           rest::binary>>,
          true
        ) do
     extension = %Header.Extension{profile_specific: profile_specific, data: data}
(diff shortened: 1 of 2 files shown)


mickel8 (changes requested):
It seems that one test doesn't pass. Btw. what do you think about adding one of packets from #23 to our unit tests? We would be safe in the future

mat-hek:
@mickel8 we already have such test, but it assumes the invalid behaviour, and this is the one failing. I've seen that and thus removed your review request ;)

lib/membrane/rtp/packet.ex
```diff
@@ -43,7 +43,13 @@ defmodule Membrane.RTP.Packet do
...
 
   defp serialize_header_extension(%Header.Extension{} = extension) do
-    <<extension.profile_specific::16, byte_size(extension.data)::16, extension.data::binary>>
+    data_size = byte_size(extension.data)
+    padded_data_size = Bunch.Math.min_multiple_gte(4, data_size)
+    padding = padded_data_size - data_size
+    length = div(padded_data_size, 4)
+
+    <<extension.profile_specific::16, length::16, extension.data::binary,
+      0::size(padding)-unit(8)>>
```
mickel8, line 52: so cool modifier

mickel8 (approved):

Issue #15, opened 2020-08-12
Support RTP/RTCP demuxing


mat-hek:
according to [RFC5761](https://tools.ietf.org/html/rfc5761)

PR #35, opened 2021-01-22
Handle inband RTCP and RTCP feedback packets

mat-hek:
This PR adds support for incoming in-band RTCP and sending RTCP receiver reports and feedback packets. Sending in-band RTCP sender reports is not supported yet.

Partially solves #15


lib/membrane/rtcp/feedback_packet/pli.ex
```diff
@@ -0,0 +1,23 @@
...
+  @impl true
+  def decode(_binary) do
+    {:ok, %__MODULE__{}}
+  end
+
+  @impl true
+  def encode(_packet) do
+    <<>>
+  end
+end
```
bblaszkow06, line 23: That actually may be a valid implementation as PLI packet has no FCI, just SSRCs common for feedback packets

lib/membrane/rtcp/feedback_packet/fir.ex
```diff
@@ -0,0 +1,36 @@
...
+    {:ok, Enum.reverse(acc)}
+  end
+
+  defp do_decode(<<target_ssrc::32, seq_num, 0::24, rest::binary>>, acc) do
+    do_decode(rest, [%__MODULE__{target_ssrc: target_ssrc, seq_num: seq_num} | acc])
+  end
+
+  defp do_decode(_binary, _acc) do
+    {:error, :malformed_packet}
+  end
```
bblaszkow06, line 35: You can use `for` for it
```suggestion
  def decode(binary) when div(bit_size(binary), 32+8+24) != 0 do
    {:error, :malformed_packet}
  end

  def decode(binary) do
    for <<target_ssrc::32, seq_num::8, 0::24 <- binary>> do
      %__MODULE__{target_ssrc: target_ssrc, seq_num: seq_num}
    end
  end
```

lib/membrane/rtcp/header.ex
```diff
@@ -26,18 +26,19 @@ defmodule Membrane.RTCP.Header do
          packet_type: pt
        },
        padding?: padding == 1,
-       length: (length + 1) * 4
+       length: length * 4
```
bblaszkow06, line 29: This was done for a reason - do you really want to keep the packet byte size - 1? This makes no sense now. Either keep a raw value of the field in RTP packet or revert that change and keep a valid byte size like it was

lib/membrane/rtcp/header.ex
```diff
@@ -26,18 +26,19 @@ defmodule Membrane.RTCP.Header do
...
     :error
   end
 
   @spec serialize(t(), length: pos_integer(), padding?: boolean()) :: binary()
   def serialize(%__MODULE__{} = header, opts) do
     padding = if Keyword.get(opts, :padding?), do: 1, else: 0
-    length = div(Keyword.fetch!(opts, :length), 4) - 1
-    <<2::2, padding::1, header.packet_specific::5, header.packet_type::8, length::16>>
+    length = Keyword.fetch!(opts, :length)
+    0 = rem(length, 4)
```
bblaszkow06, line 41: Dafuq? If you want to make a check, explicitly raise with some human readable error

lib/membrane/rtcp/receiver.ex
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  @impl true
+  def handle_tick(:fir_timer, _ctx, state) do
+    send_fir(state)
+  end
+
+  @impl true
+  def handle_event(:input, %RTCPEvent{rtcp: %SenderReportPacket{} = rtcp} = event, _ctx, state) do
+    <<_::16, cut_wallclock_ts::32, _::16>> =
```
bblaszkow06, line 59: Name the ignored values

bblaszkow06 (changes requested):

lib/membrane/rtcp/feedback_packet/fir.ex
mat-hek: `div(bit_size(binary), 32+8+24) != 0` looks ugly and is error-prone, as it requires having the sizes in two places. I'd stick to the original version if there's no way to get rid of that

lib/membrane/rtcp/header.ex
mat-hek: Here it makes the perfect sense:
- it's not stored, it's just returned
- it's needed to parse the body because the header is already parsed, so we'd have to subtract that header size in `RTCP.Packet.parse`

though I'll change that to `body_length` to make it more meaningful

lib/membrane/rtcp/feedback_packet/fir.ex
bblaszkow06: 1. Ugliness is in the eye of the beholder, I prefer the div version for stating explicitly what is considered a valid packet and what isn't
2. It's as error prone as the first version where you type sizes from RFC by hand and If you want to do this once, create module attribute. It's not like the RFC is going to change in that matter

lib/membrane/rtcp/header.ex
bblaszkow06: > it's not stored, it's just returned

It is stored in a struct under `length` key. The struct is then returned which makes no difference

> though I'll change that to body_length to make it more meaningful

Then it would make sense - it's clearly something different from the field in header, but I'd name it `body_size` as length is usually something that has to be calculated like in `Kernel.length`

lib/membrane/rtcp/feedback_packet/fir.ex
mat-hek: RFC won't change, but the code may be refactored. Keeping sizes in module attributes requires additional boilerplate, while the purpose of that change is to be more concise

bblaszkow06: >RFC won't change, but the code may be refactored

But the sizes won't

> requires additional boilerplate

If you consider `@` and name a boilerplate then you can start naming variables as `1`,  `2` to avoid boilerplate :stuck_out_tongue: 
Here's the said boilerplate: 
```
@fir_packet_size 32+8+24
...
def decode(binary) when div(bit_size(binary), @fir_packet_size) != 0
```
Besides, It's much less boilerplate than creating a recursive parsing function which requires a wrapper function, `Enum.reverse` and list appending instead of returning value

mat-hek: > But the sizes won't

But the code that contains the sizes may be.

> @fir_packet_size 32+8+24

This actually solves nothing, because you still have the sizes in two places.

mat-hek: This discussion makes no sense, as the `for` version won't fail when `reserved` part is not set to 0

bblaszkow06 (approved):

PR #40, opened 2021-04-06
Voice activity detection (vad) based on audio_level in RTP header

mickel8:


mickel8:
Do we want to include this element in `SessionBin` and add option for turning it on/off?

lib/membrane/rtp/vad.ex
```diff
@@ -0,0 +1,158 @@
...
+    {{:ok, actions}, state}
+  end
+
+  defp filter_old_audio_levels(state) do
+    Enum.reduce_while(state.audio_levels, state, fn {level, timestamp}, state ->
+      if state.current_timestamp - timestamp > state.time_window do
+        {_level, audio_levels} = Qex.pop(state.audio_levels)
+        state = %{state | audio_levels_sum: state.audio_levels_sum - level}
+        state = %{state | audio_levels_count: state.audio_levels_count - 1}
+        state = %{state | audio_levels: audio_levels}
```
mat-hek, line 98:
```suggestion
        state = %{state | audio_levels_sum: state.audio_levels_sum - level,
        audio_levels_count: state.audio_levels_count - 1,
        audio_levels: audio_levels}
```

lib/membrane/rtp/vad.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  defp maybe_notify(audio_levels_vad, state) do
+    if vad_silence?(audio_levels_vad, state) or vad_speech?(audio_levels_vad, state) do
+      [notify: {:vad, audio_levels_vad}]
+    else
+      []
+    end
+  end
+
+  defp update_state(audio_levels_vad, state) do
```
mat-hek, line 132:
```suggestion
  defp update_vad_state(audio_levels_vad, state) do
```

mat-hek:
> Do we want to include this element in `SessionBin` and add option for turning it on/off?

That would be convenient. I'd add `extensions` option to the output pad.

lib/membrane/rtp/session_bin.ex
```diff
@@ -301,6 +312,23 @@ defmodule Membrane.RTP.SessionBin do
...
+          |> to(rtp_stream_name)
+          |> to({:vad, ssrc})
+          |> to_bin_output(pad)
+        ]
+
+        {new_children, new_links}
+      else
+        {new_children, new_links}
+      end
+
```
mickel8, line 331: Tried to write macro
```elixir
defmacro add_extension_link(prev, child, enabled) do
  if enabled do
    quote do
      to(unquote(prev), unquote(child))
    end
  else
    quote do
      unquote(prev)
    end
  end
end
```

It should add link to `child` if `enabled` is true, otherwise it should just return what it got but that's not how macros work

mat-hek: The problem is that if enabled do` will be evaluated in the compile-time, so `enabled` won't be a boolean, but AST of whatever is passed as `enabled`. It may work If you just make it a function.

mat-hek (approved):

PR #41, opened 2021-05-07
Added session_bin's extensions configuration

Qizot:


lib/membrane/rtp/session_bin.ex
```diff
@@ -345,29 +348,26 @@ defmodule Membrane.RTP.SessionBin do
...
+
+        {
+          Map.merge(new_children, %{extension_id => config}),
+          new_link |> to(extension_id)
+        }
+      end)
+
+    new_links = [
+      router_link |> to_bin_output(pad)
+    ]
```
mickel8, line 370: Can we write this in one line?

lib/membrane/rtp/session_bin.ex
```diff
@@ -136,12 +139,12 @@ defmodule Membrane.RTP.SessionBin do
...
       ],
       extensions: [
-        spec: [:vad],
+        spec: [extension_t()],
         default: [],
         description: """
         List of extensions. Currently `:vad` is only supported.
         * `:vad` will turn on Voice Activity Detection mechanism firing appropriate notifications when needed.
-        Should be set only for audio tracks. For more information refer to `Membrane.RTP.VAD` module documentation.
+           Should be set only for audio tracks. For more information refer to `Membrane.RTP.VAD` module documentation.
```
mickel8, line 147: Can we check if this doesn't introduce one more tab? By running `mix docs`.

mat-hek (approved):

lib/membrane/rtp/session_bin.ex
Qizot: It seems that it does not change anything but I reverted it back.

PR #52, opened 2021-08-23
Fixed calculating timestamp_base for jitter buffer for non monotonic timestamps in RTP packets

Qizot:
As mentioned in the comment. Timestamps in RTP does not have to be monotonic regarding two consecutive packets. 

Such situation can happen e.g. with H264 when we have two frames B & P and for presentation B should be displayed before P.
 
Their timestamps will be in PTS order but sequence numbers will be incremented in DTS order which will break calculating the `timestamp_base` once such situation happens.

The following change should be break anything as the original intent was to detect timestamp overflow which in this case will be detected correctly.


lib/membrane/rtp/jitter_buffer.ex
```diff
@@ -194,8 +194,12 @@ defmodule Membrane.RTP.JitterBuffer do
     %{timestamp: rtp_timestamp} = buffer.metadata.rtp
     timestamp_base = state.timestamp_base || rtp_timestamp
 
+    # timestamps in RTP don't have to be monotonic therefore there can be
+    # a situation where in 2 consecutive packets the latter smaller timestamp
+    # then the previous packet while not overflowing timestamp number
+    # https://datatracker.ietf.org/doc/html/rfc3550#section-5.1
     timestamp_base =
-      if rtp_timestamp < state.previous_timestamp,
+      if rtp_timestamp - state.previous_timestamp < -@max_timestamp / 2,
```
mat-hek, line 202: I suppose we have another case to consider here - we can receive a packet from before the overflow. Let's assume `rtp_timestamp = @max_timestamp - 10, state.previous_timestamp = 10`. In this case, we'll generate a huge gap in timestamps, while received a packet from ~20 ticks before

mat-hek (changes requested):

lib/membrane/rtp/jitter_buffer.ex
Qizot: Right, I think we got a big problem here. Without additional information we can't know if the consecutive packets are in the right order and if their timestamps are correct, when it happens inside the range it is easy to detect, but for when we approach overflow border I guess we can't do much about it...

mat-hek (approved):

PR #55, opened 2021-08-26
RTCP packet ignores unknown packet types

sax:
In playing around with screen sharing, we've been seeing warnings about dropped RTCP packets. When we looked deeper, we found that RTCP packets with packet type `207` were being included in compound RTCP packets.

While a more complete solution would be for us to read [RFC3611](https://datatracker.ietf.org/doc/html/rfc3611) and implement Extended Report Block parsing for `207`, and possibly handle packet types `208` and `209`, we thought that a safe first step was for the plugin to ignore unknown packet types, while passing along any parts of a compound packet that were successfully parsed.

If a known packet type cannot be parsed, then the entire compound packet fails.


lib/membrane/rtcp/packet.ex
```diff
@@ -73,15 +74,38 @@ defmodule Membrane.RTCP.Packet do
...
-         {:ok, packet_module} <- BiMap.fetch(@packet_type_module, header.packet_type),
-         {:ok, packet} <- packet_module.decode(body, header.packet_specific) do
-      do_parse(rest, [packet | acc])
+         {:ok, {body, _padding}} <- RTP.Utils.strip_padding(body, padding?) do
+      case parse_packet(body, header) do
+        {:ok, packet} ->
+          do_parse(rest, [packet | acc])
+
+        {:error, :unknown_packet_type} = error ->
+          Membrane.Logger.warn("""
```
sax, line 83: This is still very noisy in Chrome. I'm considering switching this to a `debug`.

lib/membrane/rtcp/packet.ex
```diff
@@ -73,15 +74,38 @@ defmodule Membrane.RTCP.Packet do
...
+          Error parsing rtcp packet:
+          #{inspect(raw_header <> body_and_rest, limit: :infinity)}
+          Reason: #{inspect(error)}
+          """)
```
sax, line 87: This still winds up being very noisy in Chrome. I'm considering switching this to a `debug`, to make the logs a little cleaner.

lib/membrane/rtcp/packet.ex
```diff
@@ -73,15 +74,38 @@ defmodule Membrane.RTCP.Packet do
...
+         {:ok, {body, _padding}} <- RTP.Utils.strip_padding(body, padding?) do
+      case parse_packet(body, header) do
+        {:ok, packet} ->
+          do_parse(rest, [packet | acc])
+
+        {:error, :unknown_packet_type} = error ->
+          Membrane.Logger.debug("""
+          Ignoring rtcp packet with packet type #{header.packet_type}:
+          #{inspect(raw_header <> body_and_rest, limit: :infinity)}
+          Reason: #{inspect(error)}
```
Qizot, line 86: I think we should keep the reason as the error's tuple reason as we do in other places. Also we should decide if handling of such packets should have log level `debug` or `warn`, from what I see we mostly use warn for invalid packets.

Qizot (changes requested):

lib/membrane/rtcp/packet.ex
sax: One of the issues is that BiMap returns `:error` when it does not find a match, which can be pretty confusing. I tried to write it so that this specific error is caught and made more clear, but actual errors in parsing are passed through.

Qizot: Sure, that is not the problem. I just meant using `Reason: :unkown_packet_type` instead of `Reason: {:error, :unkown_packet_type}`

sax (commented):
I just pushed changes to retain the error without the `:error` tuple when printing. I also switched back to `warn`, to be more consistent with the rest of the codebase.

lib/membrane/rtcp/packet.ex
sax: Whoops. Didn't mean to mark your comments are resolved. Still new to the Github PR workflow.

Qizot (approved):

Issue #58, opened 2021-09-13
:rtp_parser input overflows when no tracks are added


sax:
I suspect that this might account for crashes when many participants join at the same time. This will definitely cause problems if there is a delay in the browser in sending tracks, but there is no delay in sending RTCP packets like sender reports.

The problem can be stated as follows:
* When a `SessionBin` is initialized, it starts the following children:
  * `RTP.Parser`
  * `RTCP.Parser`
  * `RTP.SSRCRouter`
* These children are initialized with a default demand.
* If an SSRC is identified in this initial demand, then the following happens:
  * An `RTP.StreamReceiveBin` is added to the bin, creating an output pad to the `RTP.SSRCRouter`.
  * The output of the receive bin is linked to output of the `SessionBin`, creating an output pad.
* If no SSRC is identified in the initial demand, then no output is added to the `SessionBin`.
* If the `SessionBin` does not have an output pad, then no further demand bubbles up to the `SSRCRouter`, and therefore no more demand is given to the `RTP.Parser`.

This is particularly apparent when one participant joins a room with tracks, but other participants join as passive participants. When an RTP client joins in receive-only mode, it sends RTCP receiver reports on an interval. These receiver reports eat up the initial demand, then eventually build up in the rtp parser and overflow the input buffer.

I would need to look deeper into the crashes that occur when many participants join at the same time, but if there's any possibility that the clients might send other types of RTCP packets before starting their streams, it might be triggering this same problem.

PR #59, opened 2021-09-17
Redemand outputs when parsing inputs

sax:
This solves for #58. 

* Solves for the case when there is no track sent, and a bin using this
  plugin does not ever add an output to the SessionBin
* Clients will send RTCP receiver reports when they consume tracks but do not
  send any. If there is no output on the session bin generating demand, and there
  is no redemand, then packets will eventually pile up in the RTC.Parser input
  buffer and overflow.

I can verify that this stops our input buffer overflows, but at this time I don't understand if constantly redemanding will have any adverse consequences.

Let me know what you think!


sax:
One way to keep from constantly issuing redemands would be for the session bin to forward `:session_output_added` or `:session_output_removed` messages to the rtp and rtcp parsers, and they could redemand based on keeping track of this in their states.

sax:
Seems like there's a race condition on the session output pad removed... in the context of the webrtc plugin, this happens when the endpoint is going away (the participant left the room), so by the time we forward a message to the rtp parser, it has already been shut down.

lib/membrane/rtp/parser.ex
```diff
@@ -52,7 +52,7 @@ defmodule Membrane.RTP.Parser do
 
   @impl true
   def handle_init(opts) do
-    {:ok, %{rtcp_output_pad: nil, secure?: opts.secure?}}
+    {:ok, %{rtcp_output_pad: nil, secure?: opts.secure?, has_session_output?: false}}
```
mat-hek, line 55: I suppose that `has_session_output` is set whenever `rtcp_output_pad` is not nil

lib/membrane/rtp/parser.ex
```diff
@@ -80,15 +85,27 @@ defmodule Membrane.RTP.Parser do
...
+          [buffer: {:output, %Buffer{payload: payload, metadata: metadata}}],
+          state
+        )
+
+      {{:ok, actions}, state}
     else
       :rtcp ->
         case state.rtcp_output_pad do
           nil ->
             {:ok, state}
```
mat-hek, line 99: redemand should be put here and in general whenever RTCP packet is received, as we don't send a buffer to the `output` pad in that case

lib/membrane/rtp/parser.ex
```diff
@@ -114,10 +131,15 @@ defmodule Membrane.RTP.Parser do
         {:ok, state}
 
       pad ->
-        {{:ok, event: {pad, event}}, state}
+        {{:ok, event: {pad, event}, redemand: :output}, state}
```
mat-hek, line 134: I believe we don't need to redemand here

mat-hek (changes requested):

lib/membrane/rtp/parser.ex
sax: I don't *think* this is true. In all cases I've seen so far, `rtcp?` is true, whether or not an output has been added to to the session bin.

https://github.com/membraneframework/membrane_rtp_plugin/blob/c24abf0e5480564021417e90f0e55b38aaf5acf6/lib/membrane/rtp/session_bin.ex#L267-L269
https://github.com/membraneframework/membrane_rtp_plugin/blob/c24abf0e5480564021417e90f0e55b38aaf5acf6/lib/membrane/rtp/session_bin.ex#L279-L301

lib/membrane/rtp/parser.ex
sax: I think I see what you're saying. I'll refactor so that the `with` comprehension returns a list of actions, and this `handle_process` will always check for redemand.

lib/membrane/rtp/parser.ex
sax: 👍🏻 makes sense.

lib/membrane/rtp/parser.ex
sax: I think that even if the `session_bin` does not have an output pad, the `rtp -> rtcp -> ssrc_router` links are established.

sax:
Just force pushed after rebasing from upstream, with new commits for your suggestions. Let me know if you see any other issues or desired changes!

lib/membrane/rtp/parser.ex
mat-hek: > In all cases I've seen so far, rtcp? is true, whether or not an output has been added to to the session bin.

Hmm, that's weird, because `rtcp?` seems to be set only when the rtcp output of the session bin is connected. IMHO that wouldn't work otherwise, because within `if rtcp?` we're linking to the rtcp output of the session bin.

sax:
@mat-hek notes from our conversation today:

The triangle of elements `RTP.Parser`, `RTCP.Parser`, and `SSRCRouter` makes it complicated to reason about. Although this helps to follow the RTP spec, where RTP and RTCP packets may be received from different sockets, in WebRTC all packets are generally sent over the same socket.

1. Try to simplify this by always redemanding when the `RTP.Parser` receives an rtcp packet, and removing any conditional logic.
2. See if the `RTP.Parser` and `RTCP.Parser` can be collapsed into a single element, either by taking advantage of having multiple elements running in a single process (pending the ability to do this in membrane_core), or by changing this to use module/function composition instead of processes. The `RTCP.Parser` could be a collection of functions, rather than element callbacks.

The main use case we have for this is for "passive" participants who do not send any A/V tracks, which is currently not supported by the SDP negotiation flow of the current `membrane_rtc_engine`. I'll wait a few days to watch the other repos—if the SDP negotiation updates to support passive participants don't land soon, I'll try to find another way to trigger this crash and verify that simpler code fixes it.

PR #63, opened 2021-10-20
Migration to buffer with pts dts

andpodob:
Updates plugin to use `:pts` and `:dts` fields in `Membrane.Buffer`


lib/membrane/rtp/serializer.ex
```diff
@@ -85,10 +85,11 @@ defmodule Membrane.RTP.Serializer do
       extension: Map.get(rtp_metadata, :extension)
     }
 
-    packet = %RTP.Packet{header: header, payload: payload}
+    packet = %RTP.Packet{header: header, payload: buffer.payload}
     payload = RTP.Packet.serialize(packet, align_to: state.alignment)
 
     buffer = %Buffer{
+      pts: Buffer.get_dts_or_pts(buffer),
```
mat-hek, line 92: Here we potentially set dts as pts, and set dts to nil... why?

andpodob: Yes, you're right, that's probably messed up, I need to rethink that a bit.

Qizot (approved):

lib/membrane/rtp/serializer.ex
```diff
@@ -77,11 +77,11 @@ defmodule Membrane.RTP.Serializer do
...
   @impl true
-  def handle_process(:input, %Buffer{payload: payload, metadata: metadata}, _ctx, state) do
-    {rtp_metadata, metadata} = Map.pop(metadata, :rtp, %{})
+  def handle_process(:input, buffer, _ctx, state) do
+    {rtp_metadata, metadata} = Map.pop(buffer.metadata, :rtp, %{})
 
     rtp_offset =
-      metadata.timestamp
+      buffer.pts
       |> Ratio.mult(state.clock_rate)
```
mat-hek, line 85: ratio seems not needed

mat-hek (approved):

mat-hek:
also Ratio not needed in VAD and jitterbuffer/timestamp_calculation_test.exs

PR #70, opened 2021-11-09
Handle extensions

Rados13:
Implement the mechanism described in [RFC 8285](https://datatracker.ietf.org/doc/html/rfc8285).  Support only a one-byte header, as according to RFC this form is preferred and it must be supported by all receivers.

### DIFF
--- a/lib/membrane/rtp/header_extension.ex
+++ b/lib/membrane/rtp/header_extension.ex
@@ -17,7 +17,7 @@ defmodule Membrane.RTP.Header.Extension do
   defstruct @enforce_keys
 
   @type t :: %__MODULE__{
-          identifier: binary(),
+          identifier: 1..14,
           data: binary()
         }
 end
--- a/lib/membrane/rtp/header_extension.ex
+++ b/lib/membrane/rtp/header_extension.ex
@@ -13,11 +13,11 @@ defmodule Membrane.RTP.Header.Extension do
 
   ```
   """
-  @enforce_keys [:profile_specific, :data]
+  @enforce_keys [:identifier, :data]
   defstruct @enforce_keys
 
   @type t :: %__MODULE__{
-          profile_specific: binary(),
+          identifier: binary(),
           data: binary()
         }
 end
--- a/lib/membrane/rtp/packet.ex
+++ b/lib/membrane/rtp/packet.ex
@@ -14,6 +14,9 @@ defmodule Membrane.RTP.Packet do
   @enforce_keys [:header, :payload]
   defstruct @enforce_keys
 
+  # 0xBEDE for Venerable Bede from RFC 8285
+  @one_byte_header_identifier <<190, 222>>
+
   @spec identify(binary()) :: :rtp | :rtcp
   def identify(<<_first_byte, _marker::1, payload_type::7, _rest::binary>>)
       when payload_type in 64..95,
@@ -26,14 +29,14 @@ defmodule Membrane.RTP.Packet do
     %__MODULE__{header: header, payload: payload} = packet
     %Header{version: 2} = header
     has_padding = 0
-    has_extension = if header.extension, do: 1, else: 0
+    has_extension = if header.extensions, do: 1, else: 0
     marker = if header.marker, do: 1, else: 0
     csrcs = Enum.map_join(header.csrcs, &<<&1::32>>)
 
     serialized =
       <<header.version::2, has_padding::1, has_extension::1, length(header.csrcs)::4, marker::1,
         header.payload_type::7, header.sequence_number::16, header.timestamp::32, header.ssrc::32,
-        csrcs::binary, serialize_header_extension(header.extension)::binary, payload::binary>>
+        csrcs::binary, serialize_header_extensions(header.extensions)::binary, payload::binary>>
 
     case Utils.align(serialized, align_to) do
       {serialized, 0} ->
@@ -43,6 +46,17 @@ defmodule Membrane.RTP.Packet do
         <<pre::2, _has_padding::1, post::bitstring>> = serialized
         <<pre::2, 1::1, post::bitstring>>
     end
+  end
+
+  defp serialize_header_extensions([]), do: <<>>
+
+  defp serialize_header_extensions(extensions) do
+    extensions =
+      Enum.reduce(extensions, <<>>, fn extension, acc ->
+        acc <> serialize_header_extension(extension)
+      end)
+
+    <<@one_byte_header_identifier, byte_size(extensions) |> div(4)::16>> <> extensions
   end
 
   defp serialize_header_extension(nil) do
@@ -68,7 +82,7 @@ defmodule Membrane.RTP.Packet do
           rest::binary>> = original_packet,
         encrypted?
       ) do
-    with {:ok, {extension, payload}} <-
+    with {:ok, {extensions, payload}} <-
            parse_header_extension(rest, has_extension == 1),
          {:ok, {payload, padding}} =
            Utils.strip_padding(payload, not encrypted? and has_padding == 1) do
@@ -80,7 +94,7 @@ defmodule Membrane.RTP.Packet do
         payload_type: payload_type,
         timestamp: timestamp,
         csrcs: for(<<csrc::32 <- csrcs>>, do: csrc),
-        extension: extension
+        extensions: extensions
       }
 
       {:ok,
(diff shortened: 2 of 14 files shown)


lib/membrane/rtp/packet.ex
```diff
@@ -14,6 +14,9 @@ defmodule Membrane.RTP.Packet do
   @enforce_keys [:header, :payload]
   defstruct @enforce_keys
 
+  # 0xBEDE for Venerable Bede from RFC 8285
+  @one_byte_header_identifier <<190::8, 222::8>>
```
Qizot, line 18: Can you simply use hex values? I use that inside `Membrane.RTP.SilenceDiscarder`.

lib/membrane/rtp/packet.ex
```diff
@@ -45,13 +48,37 @@ defmodule Membrane.RTP.Packet do
...
+
+    identifer = @one_byte_header_identifier
+    padding = calculate_padding_size(byte_size(extensions)) * 8
+    extensions_size = byte_size(extensions)
+
+    extension_header_size =
+      div(extensions_size, 4) + if rem(extensions_size, 4) == 0, do: 0, else: 1
+
+    <<identifer::binary-size(2), extension_header_size::16>> <>
+      extensions <> <<0::integer-size(padding)>>
```
Qizot, line 67: can you put that all into a single binary? Erlang then can optimize the call and right now it will probably create 3 diferent binaries and concatenate them.

lib/membrane/rtp/packet.ex
```diff
@@ -45,13 +48,37 @@ defmodule Membrane.RTP.Packet do
...
+      end)
+
+    identifer = @one_byte_header_identifier
+    padding = calculate_padding_size(byte_size(extensions)) * 8
+    extensions_size = byte_size(extensions)
+
+    extension_header_size =
+      div(extensions_size, 4) + if rem(extensions_size, 4) == 0, do: 0, else: 1
+
+    <<identifer::binary-size(2), extension_header_size::16>> <>
```
Qizot, line 66: why don't you put the module attribute here instead of reassigning it to identiier?

test/membrane/rtp/packet_test.exs
```diff
@@ -49,12 +49,11 @@ defmodule Membrane.RTP.PacketTest do
   end
 
   test "reads and serializes extension header" do
-    extension_header = <<0::16, 4::16, 1::32, 2::32, 3::32, 4::32>>
+    extension_header = <<190::8, 222::8, 1::16, 1::4, 0::4, 190::8, 0::16>>
```
Qizot, line 52: can you use hex when using numbers > 10?

Qizot (changes requested):

lib/membrane/rtp/packet.ex
```diff
@@ -100,17 +126,38 @@ defmodule Membrane.RTP.Packet do
...
+
+  defp parse_one_byte_header(<<extension_header::bits-size(8), extensions::binary>>) do
+    <<local_identifier::integer-size(4), len_data::integer-size(4)>> = extension_header
+
+    if local_identifier == 0 do
+      {:ok, extensions}
+    else
+      len_data = len_data + 1
+      <<data::binary-size(len_data), next_extensions::binary>> = extensions
+      extension = %Header.Extension{profile_specific: local_identifier, data: data}
```
balins, line 159: `Header.Extension has now wrong typespec 👀 
```elixir
@type t :: %__MODULE__{
          profile_specific: binary(),
          data: binary()
        }
```
`profile_specific` in RFC 8285 are used to distinguish between one- and two-byte headers. As we target one-byte headers for now, I think we should consider replacing it with `identifier` field.

lib/membrane/rtp/serializer.ex
```diff
@@ -84,7 +84,7 @@ defmodule Membrane.RTP.Serializer do
       timestamp: rtp_timestamp,
       sequence_number: state.sequence_number,
       csrcs: Map.get(rtp_metadata, :csrcs, []),
-      extension: Map.get(rtp_metadata, :extension)
+      extensions: [Map.get(rtp_metadata, :extension)]
```
Qizot, line 87: Will this need further refactor in other payloaders to return the list or are we fine with returing just one extension from them?

test/membrane/rtp/packet_test.exs
```diff
@@ -49,12 +49,11 @@ defmodule Membrane.RTP.PacketTest do
   end
 
   test "reads and serializes extension header" do
-    extension_header = <<0::16, 4::16, 1::32, 2::32, 3::32, 4::32>>
+    extension_header = <<0xBE::8, 0xDE::8, 1::16, 1::4, 0::4, 0xBE::8, 0::16>>
```
Qizot, line 52: I don't think you need to specify the number of pits for the hex numbers :P

lib/membrane/rtp/packet.ex
```diff
@@ -100,17 +126,39 @@ defmodule Membrane.RTP.Packet do
...
+  defp parse_data_while_possible(<<>>, acc), do: acc
+
+  defp parse_data_while_possible(data, acc) do
+    case parse_one_byte_header(data) do
+      {:ok, {data, rest}} -> parse_data_while_possible(rest, [data | acc])
+      {:ok, rest} -> parse_data_while_possible(rest, acc)
+    end
+  end
+
+  defp parse_one_byte_header(<<extension_header::bits-size(8), extensions::binary>>) do
```
Qizot, line 152: An explanation why we don't support two byte header would be appreciated :)

Qizot (changes requested):

lib/membrane/rtp/packet.ex
```diff
@@ -100,17 +126,39 @@ defmodule Membrane.RTP.Packet do
...
+    <<local_identifier::integer-size(4), len_data::integer-size(4)>> = extension_header
+
+    if local_identifier == 0 do
+      {:ok, extensions}
```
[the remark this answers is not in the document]
Rados13: There was, but now this is not useful so I will change it.

lib/membrane/rtp/serializer.ex
Rados13: This was a bug, It will look like this now 
```elixir
extensions: Map.get(rtp_metadata, :extensions)
```

lib/membrane/rtp/packet.ex
Rados13: According to RFC 8285 section 4.1.2 ", `the one-byte header form is preferred and MUST be supported by all receivers"`. When I was implementing this, I did not need two-byte headers. But I can add it to this PR.

Qizot (changes requested):

lib/membrane/rtp/packet.ex
```diff
@@ -100,17 +129,39 @@ defmodule Membrane.RTP.Packet do
...
+  defp parse_extension_data(<<>>, acc), do: acc
+
+  defp parse_extension_data(data, acc) do
+    case parse_one_byte_header(data) do
+      {data, rest} -> parse_extension_data(rest, [data | acc])
+      rest -> parse_extension_data(rest, acc)
+    end
+  end
+
+  defp parse_one_byte_header(<<extension_header::bits-size(8), extensions::binary>>) do
```
Qizot, line 155: Could you do that `if` statement as two different function pattern matches as you are already matching on it?

Qizot (changes requested):

lib/membrane/rtp/packet.ex
```diff
@@ -152,16 +152,17 @@ defmodule Membrane.RTP.Packet do
     end
   end
 
-  defp parse_one_byte_header(<<extension_header::bits-size(8), extensions::binary>>) do
-    <<local_identifier::integer-size(4), len_data::integer-size(4)>> = extension_header
+  defp parse_one_byte_header(
```
Qizot, line 155: This is much better :)

Qizot (approved):

lib/membrane/rtp/header_extension.ex
```diff
@@ -13,11 +13,11 @@ defmodule Membrane.RTP.Header.Extension do
 
   ```
   """
-  @enforce_keys [:profile_specific, :data]
+  @enforce_keys [:identifier, :data]
   defstruct @enforce_keys
 
   @type t :: %__MODULE__{
-          profile_specific: binary(),
+          identifier: binary(),
```
balins, line 20: It's not binary anymore 😉

balins:
I think we should also update VAD to reflect these changes

lib/membrane/rtp/header_extension.ex
balins, line 20:
```suggestion
          identifier: pos_integer(),
```

balins, line 20: Or maybe to be more strict:
```suggestion
          identifier: 1..14,
```

balins (approved):

PR #71, opened 2021-11-09
Support TWCC

### LINKED ISSUE #16: Support WebRTC custom RTCP messages support (REMB, transport-wide-cc)
From what I've learned from [this great talk](https://gstconf.ubicast.tv/videos/google-transport-wide-congestion-control/) about the [TWCC implementation in GStreamer](https://github.com/pexip/gst-plugins-good/blob/3c0d7bc94fc379857ddbfa3e85b14dd9531c76b9/gst/rtpmanager/rtptwcc.c):
- both REMB and [TWCC](https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01) are congestion control mechanisms, so WebRTC can work without them when the link is good
- TWCC seems to be a successor of REMB, so there's probably no point in supporting REMB (chrome, janus and mediasoup support TWCC, not sure how about firefox and other browsers)
- basically, TWCC contains all the information about the presence and delays between packets
- the receiver only needs to generate the packets, while the sender has to adapt some algorithm to adjust the bandwidth (the simplest one would be to calculate the bandwidth from TWCC and cap to that bandwidth)
- as the guy said in the talk, Chrome fails to handle TWCC packets that are compound with other RTCP packets (eg receiver reports)

balins:
Adds support for receiver side of [TWCC](https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01) (gathering transport-wide packet stats, sending feedback packets).

related to #16

pipelines-graph

### Changes

The PR adds `RTP.TWCCReceiver` element inserted between `RTP.SSRCRouter` and possibly many `StreamReceiveBin`s. The element gathers information required to create TWCC feedback packets, sending them at a fixed interval.

Other changes include adding a new type of RTCP packet: `RTCP.TransportFeedbackPacket` that represent [RFC 4585 transport layer feedback messages](https://datatracker.ietf.org/doc/html/rfc4585#section-6.2) along with a module for parsing/serializing TWCC statistics (`RTCP.TransportFeedbackPacket.TWCC`).

### Testing 
To try it out you need to use `transport_cc` branch of [membrane_webrtc_plugin](https://github.com/membraneframework/membrane_webrtc_plugin/tree/transport_cc), overriding dependency in `mix.exs` of videoroom demo/RTC Engine. To enable TWCC, pass `webrtc_extensions: [Membrane.WebRTC.Extension.TWCC]` as RTC Engine option.


lib/membrane/rtp/session_bin.ex
```diff
@@ -375,9 +384,31 @@ defmodule Membrane.RTP.SessionBin do
...
+      else
+        {new_children, state}
+      end
+
+    maybe_link_twcc =
+      &(&1
+        |> via_in(Pad.ref(:input, ssrc))
+        |> to(:twcc)
+        |> via_out(Pad.ref(:output, ssrc)))
+
```
Qizot, line 407: Can we put this whole logic to a separate function and just use it in pipe chain? Should be more readable.

lib/membrane/rtp/session_bin.ex
```diff
@@ -375,9 +384,31 @@ defmodule Membrane.RTP.SessionBin do
...
 
+    # workaround: as TWCC is a transport-wide extension, there should exist only one TWCC child
+    # that handles packets from all incoming streams that have declared support for it
+    {maybe_twcc, rtp_extensions} = Keyword.pop(rtp_extensions, :twcc)
+
+    should_link_twcc? = maybe_twcc != nil
+
+    {new_children, state} =
+      if should_link_twcc? and not Map.has_key?(ctx.children, :twcc) do
+        {twcc_ssrc, state} = add_ssrc(nil, state)
```
Qizot, line 395: Any specific reason for putting `nil` ssrc for the twcc in the state? Why do we need that mapping in the end?

lib/membrane/rtp/twcc_receiver.ex
```diff
@@ -0,0 +1,155 @@
...
+                default: nil,
+                description:
+                  "Sender SSRC for generated feedback packets (will be supplied by `RTP.SessionBin`)."
+              ]
+
+  defmodule State do
+    @moduledoc false
+    alias Membrane.RTP.TWCCReceiver.PacketInfoStore
+
+    @type t :: %__MODULE__{
```
Qizot, line 37: I don't think that this typespec is correct (unless RTP.ssrc_t() specifies that it can be nil). Both of your `sender/media` ssrcs can be nil.

lib/membrane/rtp/twcc_receiver/packet_info_store.ex
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  defp maybe_handle_rollover(store, new_seq_num) do
+    %{
+      base_seq_num: base_seq_num,
+      max_seq_num: max_seq_num,
+      seq_to_timestamp: seq_to_timestamp
+    } = store
+
+    case from_which_cycle(base_seq_num, new_seq_num) do
```
Qizot, line 67: Is this the same logic as in jitter buffer element? Right now this logic is spreaded in 3 places, JitterBuffer, VAD and now here. Can we extract it to one place (only if this logic is common for all of the modules)?

test/membrane/rtp/twcc_receiver/packet_info_store_test.exs
```diff
@@ -0,0 +1,195 @@
+defmodule Membrane.RTP.TWCCReceiver.PacketInfoStoreTest do
+  use ExUnit.Case, async: true
+  use Bunch
+
+  alias Membrane.RTP.TWCCReceiver.PacketInfoStore
+
+  @max_seq_number 65_535
```
Qizot, line 7: can you use bitwise like in other places?

Qizot (changes requested):

lib/membrane/rtp/session_bin.ex
balins: I needed a way to generate an SSRC for TWCC using mechanism that is already implemented (using SSRC generator passed as a `SessionBin` option and adding this SSRC to set of our local SSRCs). Unfortunately, this mechanism assumes that for every "local SSRC" (a one that is generated by `SessionBin`s generator) there is some "remote SSRC", what doesn't hold for TWCC (and probably for other transport-wide feedbacks too). As the first argument of `add_ssrc/2` is "remote SSRC", I pass `nil` to indicate that there is no remote SSRC associated with the local SSRC that will be generated.

lib/membrane/rtp/twcc_receiver.ex
balins: Fixed

lib/membrane/rtcp/transport_feedback_packet/twcc.ex
```diff
@@ -0,0 +1,221 @@
...
+
+  defp run_length_to_status_vector(%RunLength{packet_status: status, packet_count: count}),
+    do: %StatusVector{vector: Enum.map(1..count, fn _i -> status end), packet_count: count}
+
+  defp status_vector_to_run_length(%StatusVector{vector: vector}) do
+    vector
+    |> Enum.reduce([], fn status, acc ->
+      case acc do
+        [%RunLength{packet_status: ^status, packet_count: count} | rest] ->
+          [%RunLength{packet_status: status, packet_count: count + 1} | rest]
```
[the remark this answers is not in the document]
balins: Done

lib/membrane/rtp/twcc_receiver/packet_info_store.ex
balins: Good catch, in fact this logic was in 4 modules 😆 
Extracted to `RTP.Utils.from_which_cycle/3`

lib/membrane/rtp/jitter_buffer/buffer_store.ex
```diff
@@ -72,10 +72,12 @@ defmodule Membrane.RTP.JitterBuffer.BufferStore do
          buffer,
          seq_num
        ) do
+    base_index = rem(prev_index, @seq_number_limit)
```
Qizot, line 75: If you do the reminder with `@seq_number_limit` then it is no longer an index. Index is a counter that cummulates sequence numbers with their respective rollovers (while sequence number is 16 bits then the index can be much larger containing multiple sequence number maximum valu). In this case just name it `prev_seq_num`.

lib/membrane/rtp/session_bin.ex
```diff
@@ -588,4 +601,33 @@ defmodule Membrane.RTP.SessionBin do
...
+        {Map.put(new_children, :twcc, %{maybe_twcc | sender_ssrc: twcc_ssrc}), state}
+      else
+        {new_children, state}
+      end
+
+    maybe_link_twcc =
+      if should_link_twcc? do
+        &(&1
+          |> via_in(Pad.ref(:input, pad_ssrc))
+          |> to(:twcc)
```
Qizot, line 625: Instead of doing the whole `new_children` and updating the state you could simply put that element in `to(:twcc, %{maybe_twcc | sender_ssrc: twcc_ssrc})` keeping the `new_children` out of this funciton, and just update the state respectively if you need to add the `ssrc`.

lib/membrane/rtp/utils.ex
```diff
@@ -29,4 +29,24 @@ defmodule Membrane.RTP.Utils do
         {<<payload::binary, 0::size(zeros_no)-unit(8), padding_size>>, padding_size}
     end
   end
+
+  @spec from_which_cycle(number() | nil, number(), number()) :: :current | :previous | :next
+  def from_which_cycle(nil, _new, _cycle_length), do: :current
```
Qizot, line 34: can you make a function header without definition that will handle the function spec? Something like this:
```elixir
@spec from_which_cycle(number() | nil, number(), number()) :: :current | :previous | :next
def from_which_cycle(previous_value, new_value, cycle_length)
```

And after that you can do the patter matching on the function. Right now it is difficult to read the function definition when the first pattern match is `nil` and there is no further information until I look at the next match what the first param really is. And it is no necessarily `base` :)

lib/membrane/rtp/vad.ex
```diff
@@ -121,6 +121,8 @@ defmodule Membrane.RTP.VAD do
 
   defp handle_if_present(buffer, nil, state), do: {{:ok, buffer: {:output, buffer}}, state}
 
+  @timestamp_limit 4_294_967_295
```
Qizot, line 124: Can you replace it with `Bitwise` once again? It is much easier to read when we use that :)

Qizot (changes requested):

lib/membrane/rtp/session_bin.ex
balins: Done

lib/membrane/rtp/utils.ex
balins: Done

lib/membrane/rtp/session_bin.ex
```diff
@@ -588,4 +600,36 @@ defmodule Membrane.RTP.SessionBin do
...
+      else
+        {nil, state}
+      end
+
+    maybe_link_twcc =
+      if should_link? do
+        &(&1
+          |> via_in(Pad.ref(:input, pad_ssrc))
+          |> then(fn link ->
+            if should_create_child?,
```
Qizot, line 624: Normal `do/end` would be better than the shorter versions with `:` :P

lib/membrane/rtp/vad.ex
```diff
@@ -121,6 +121,8 @@ defmodule Membrane.RTP.VAD do
 
   defp handle_if_present(buffer, nil, state), do: {{:ok, buffer: {:output, buffer}}, state}
 
+  @timestamp_limit Bitwise.bsl(1, 32)
```
Qizot, line 124:
```suggestion
  @timestamp_limit Bitwise.bsl(1, 32) - 1
```

Qizot, line 124:
```suggestion
  @timestamp_limit Bitwise.bsl(1, 32) - 1
```

And also make sure that you have applied the subtraction in each place correctly.

Qizot (approved):
Just 2 comments left, fix them and other than that LGTM!

lib/membrane/rtp/session_bin.ex
balins: It's arguable, but will change 😛

lib/membrane/rtp/vad.ex
balins: I know it was like this before, but it was wrong. We should calculate to which epoch a number belongs using the length of the cycle, not the maximum value. Take a simple example, like cycle 0..3 and try it with `from_which_cycle/3` implementation.

PR #72, opened 2021-11-09
VAD finds timestamp in rtp metadata

sax:
Timestamps were removed from top level metadata. When depayloaders are removed from the bin, this would blow up for us.

Co-authored-by: Austin Putman <austinp@geometer.io>
Co-authored-by: Jason Carter <jason@geometer.io>

### DIFF
--- a/test/membrane/rtp/vad_test.exs
+++ b/test/membrane/rtp/vad_test.exs
@@ -99,6 +99,11 @@ defmodule Membrane.RTP.VADTest do
     end)
   end
 
+  defp process_buffer(buffer, state) do
+    {{:ok, _actions}, new_state} = VAD.handle_process(:input, buffer, %{}, state)
+    new_state
+  end
+
   describe "handle_process" do
     setup [
       :setup_vad_options,
@@ -116,7 +121,7 @@ defmodule Membrane.RTP.VADTest do
 
       buffer = rtp_buffer(-127, initial_timestamp)
 
-      {{:ok, [buffer: _]}, new_state} = VAD.handle_process(:input, buffer, %{}, state)
+      new_state = process_buffer(buffer, state)
 
       assert %{
                audio_levels_count: 1,
@@ -125,8 +130,7 @@ defmodule Membrane.RTP.VADTest do
              } = new_state
 
       buffer = rtp_buffer(-50, initial_timestamp + time_delta)
-
-      {{:ok, [buffer: _]}, new_state} = VAD.handle_process(:input, buffer, %{}, new_state)
+      new_state = process_buffer(buffer, new_state)
 
       assert %{
                audio_levels_count: 2,
@@ -142,8 +146,7 @@ defmodule Membrane.RTP.VADTest do
         initial_timestamp: initial_timestamp
       } = ctx
 
-      {{:ok, [buffer: _]}, state} =
-        VAD.handle_process(:input, rtp_buffer(-127, initial_timestamp), %{}, state)
+      state = process_buffer(rtp_buffer(-127, initial_timestamp), state)
 
       assert %{vad: :silence} = state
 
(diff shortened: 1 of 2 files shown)


Qizot:
The problem with this vad extension is that simply using RTP timestamp makes all those calculations invalid as those expect it to be in nanoseconds (base unit of time for membrane). As far as I know there is some ongoing work for refactoring those timestamps so as for now I would suggest turning the vad extension off.

sax:
@Qizot This latest commit takes the clock rate into consideration, and seems to fix VAD... especially as the age of the element gets into minutes. We were finding that with the old time comparison math, VAD would work... except that the Qex queue would never get cleared out, and eventually build up into so many thousands of items that it would get too slow to process.

This *works* because the default `48000` clock rate matches up with OPUS. Another possibility is to look up clock rate by payload type, maybe based on the first packet that comes through?

sax (commented):
`rtp_timestamp_increment` is not the more clear description, but we couldn't think of a better one. Basically, it's the `time_window` normalized to the difference in RTP timestamps between samples at the encoding's clock rate. At `48kHz` for OPUS, the RTP timestamp delta between packets is `960`.

test/membrane/rtp/vad_test.exs
```diff
@@ -0,0 +1,113 @@
...
+      {{:ok, _}, new_state} = VAD.handle_process(:input, buffer, %{}, state)
+      new_state
+    end)
+  end
+
+  describe "handle_process" do
+    setup :setup_vad_state
+
+    test "keeps running totals for buffers within a specific time window", %{state: state} do
+      buffer = rtp_buffer(-127, 1_201_851_607)
```
Qizot, line 72: can we start with some predefined timestamp attribute and then increment it by 960 (and some comment would be usefull why is that) by hand so that we know where those timestamps come from and how they get incremented?

Qizot: same goes for the test above where the interval is 20ms, at the moment we know where it comes from but when somebody else will look at it then it will look like sorcery to him :)

test/membrane/rtp/vad_test.exs
```diff
@@ -0,0 +1,113 @@
+defmodule Membrane.RTP.VADTest do
+  use ExUnit.Case
+
+  alias Membrane.RTP.VAD
+
+  ExUnit.Case.register_attribute(__MODULE__, :clock_rate)
```
Qizot, line 6: Do we need all these ExUnit attributes functionality if in fact we are not using it? Or am I missing something? To be honest I see this for the first time and seems super useful but I guess not in this case?

Qizot (changes requested):
Just small changes, I really like the idea of operating on rtp timestamps, seems more clear right now :)

test/membrane/rtp/vad_test.exs
sax: For one thing, I wanted to introduce this since I was pretty sure it was new to others.

I would like to write other tests that actually override the defaults and verify the outcomes... unfortunately I still don't quite understand the code enough to write those tests.

test/membrane/rtp/vad_test.exs
sax: Take a look at the changes I just pushed. It still does math in the test, but hopefully it's a little more clear than before.

test/membrane/rtp/vad_test.exs
sax: Ok, I just managed to write a few more tests that both exercise these test attributes, and cover a few more aspects of the module.

sax:
@Qizot I just added a skipped test to account for RTP timestamp rollovers. Maybe we could take a look at it together sometime next week—I'm not sure if it defines the expected behavior for how VAD should handle a rollover, and I suspect you might have a better idea than me for how to implement it.

Qizot (commented):
I really like the descriptions right now and the usage of the attributes now make a lot of sense :)

When it comes to over roll of timestamps, we should probably keep a timestamp accumulator which will be kept for the whole lifetime of element, added to each timestamp and incremented when we detect the overroll (the detection phase is not that difficult, probably very close to detecting the sequence number overroll which we already do in some elements). We can pair on that if you want to or I can give you some more hints?

Qizot (approved):

Qizot:
:partying_face:

PR #73, opened 2021-11-10
Fixed invalid timestamps calculations inside rtp serializer

Qizot:
During refactor of payloaders/depaylaoders it happened that for audio packets a wrong timestamp was being taken into rtp timestamp calculations. Right now when serializer is being used only for depayloaded streams it should take only `metadata` timestamps into consideration and ignore any `rtp` timestamps.

Before this change the serializer worked as expected for video as there was no rtp timestamp present so it always took the metadata one.


mat-hek (approved):

bblaszkow06:
cc: @andpodob, might be important when migrating to new timestamps

PR #75, opened 2021-11-10
Fix state creation on init of RTP.Serializer

bblaszkow06:
The state of the serializer was initialized incorrectly:
* The `State` struct was missing keys added in element options
* state returned from init was created using `Map.merge` thus allowing to create invalid struct i.e. a map with `__struct__` key but not recognized as the actual struct


Qizot (approved):
LGTM, I don't know why it was this way in the first place but I didn't bother changing it ¯\_(ツ)_/¯

PR #77, opened 2021-11-17
Fix handling of first VAD timestamp

sax:
When the initial RTP timestamp is very far from 0, it is thought to be from the previous epoch of 32bit integers
after an integer wrap around. When we do not have any saved audio levels, we know we should start processing.

### DIFF
--- a/lib/membrane/rtp/vad.ex
+++ b/lib/membrane/rtp/vad.ex
@@ -116,21 +116,28 @@ defmodule Membrane.RTP.VAD do
 
     cond do
       epoch == :current && rtp_timestamp > state.current_timestamp ->
-        state = %{state | current_timestamp: rtp_timestamp}
-        state = filter_old_audio_levels(state)
-        state = add_new_audio_level(state, level)
-        audio_levels_vad = get_audio_levels_vad(state)
-        actions = [buffer: {:output, buffer}] ++ maybe_notify(audio_levels_vad, state)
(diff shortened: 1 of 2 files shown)


sax:
@Qizot Changing the maximum random timestamp used by the test makes it so that the existing tests sometimes fail. I'm guessing this might be where property tests would be useful, but I'm not familiar with writing those.

Qizot:
What do you think about initializing `current_timestamp` in state to `nil` and then in `handle_process` automiatically do something like this `current_timestamp = state.current_timestmap || rtp_timestamp`.

With this approach when we are handling the first buffer the epoch will indicate the `current` and then we can probably change `>` to `>=` and make the first `cond` statement pass as we will have the same timestamp?

sax:
Particularly with RED starting to come through, I'd be worried about changing the condition to `>=`, as it could needlessly add a lot of things to the queue. I do like the idea of starting `nil`, though—the epoch function can special case it to be `:current`, and remove one of the conditions.

Qizot (approved):

Qizot:
> Particularly with RED starting to come through, I'd be worried about changing the condition to `>=`, as it could needlessly add a lot of things to the queue. I do like the idea of starting `nil`, though—the epoch function can special case it to be `:current`, and remove one of the conditions.

Yup, makes sense with the RED packets. Looks great right now 🥳

PR #82, opened 2021-11-26
Add support for handling multiple extension ids

balins:
related to membraneframework/membrane_webrtc_plugin#56

The proposed flow for multiple extensions is:
1. When adding new `:rtp_input` pad to `SessionBin` we receive definitions of enabled extensions as a pad option:
```{extension_name :: extension_name_t(), extension_config :: Membrane.ParentSpec.child_spec_t()}```
Extensions should be already initialized with their header extension IDs, so they know which extension in list of `RTP.Header.Extensions` is the one they should use, e.g. `%Membrane.RTP.VAD{vad_id: 1}`.
These elements will be responsible for replacing these integer ids with their names, e.g. `:vad` when forwarding buffers. 
2. When adding new `:rtp_output` pad to `SessionBin` we receive a mapping from extension names to header extension IDs expected by a receiver, ex. `%{vad: 14}`. This mapping is used by `OutboundPacketTracker` to put back integer IDs into header extensions in outgoing buffers.

### DIFF
--- a/lib/membrane/rtp/header_extension.ex
+++ b/lib/membrane/rtp/header_extension.ex
@@ -26,8 +26,8 @@ defmodule Membrane.RTP.Header.Extension do
           data: binary()
         }
 
-  @spec pop_identifier(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
-  def pop_identifier(buffer, identifier) do
+  @spec pop(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
+  def pop(buffer, identifier) do
     extension = Enum.find(buffer.metadata.rtp.extensions, &(&1.identifier == identifier))
 
     if extension do
--- a/lib/membrane/rtp/header_extension.ex
+++ b/lib/membrane/rtp/header_extension.ex
@@ -19,15 +19,15 @@ defmodule Membrane.RTP.Header.Extension do
   @enforce_keys [:identifier, :data]
   defstruct @enforce_keys
 
-  @type identifier_t :: 1..14 | SessionBin.extension_name_t()
+  @type identifier_t :: 1..14 | SessionBin.rtp_extension_name_t()
 
   @type t :: %__MODULE__{
           identifier: identifier_t(),
           data: binary()
         }
 
-  @spec pop(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
-  def pop(buffer, identifier) do
+  @spec pop_identifier(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
+  def pop_identifier(buffer, identifier) do
     extension = Enum.find(buffer.metadata.rtp.extensions, &(&1.identifier == identifier))
 
     if extension do
--- a/lib/membrane/rtp/header_extension.ex
+++ b/lib/membrane/rtp/header_extension.ex
@@ -13,11 +13,43 @@ defmodule Membrane.RTP.Header.Extension do
 
   ```
   """
+  alias Membrane.Buffer
+  alias Membrane.RTP.SessionBin
+
   @enforce_keys [:identifier, :data]
   defstruct @enforce_keys
 
+  @type identifier_t :: 1..14 | SessionBin.extension_name_t()
+
   @type t :: %__MODULE__{
-          identifier: 1..14,
+          identifier: identifier_t(),
           data: binary()
         }
+
+  @spec pop(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
+  def pop(buffer, identifier) do
+    extension = Enum.find(buffer.metadata.rtp.extensions, &(&1.identifier == identifier))
+
+    if extension do
+      buffer =
+        Bunch.Struct.update_in(
+          buffer,
+          [:metadata, :rtp, :extensions],
+          &delete(&1, identifier)
+        )
+
+      {extension, buffer}
+    else
+      {nil, buffer}
+    end
+  end
+
+  @spec put(Buffer.t(), t()) :: Buffer.t()
+  def put(buffer, extension) do
+    Bunch.Struct.update_in(buffer, [:metadata, :rtp, :extensions], &[extension | &1])
+  end
+
+  defp delete(extensions, identifier) do
+    Enum.reject(extensions, &(&1.identifier == identifier))
+  end
 end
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: lib/membrane/rtp/header_extension.ex [hunk 1]
```diff
@@ -13,11 +13,43 @@ defmodule Membrane.RTP.Header.Extension do
...
+  @type identifier_t :: 1..14 | SessionBin.extension_name_t()
+
   @type t :: %__MODULE__{
-          identifier: 1..14,
+          identifier: identifier_t(),
           data: binary()
         }
+
+  @spec pop(Buffer.t(), identifier_t()) :: {t() | nil, Buffer.t()}
+  def pop(buffer, identifier) do
```
Qizot, line 30:
```suggestion
  def pop_identifier(buffer, identifier) do
```

lib/membrane/rtp/outbound_packet_tracker.ex
```diff
@@ -65,8 +67,20 @@ defmodule Membrane.RTP.OutboundPacketTracker do
 
     {rtp_metadata, metadata} = Map.pop(buffer.metadata, :rtp, %{})
 
+    extensions =
```
Qizot, line 70: This whole statement seems like a lot of work that has to be done for each packet, please think how to optimize it a little bit, like `Map.keys`.

lib/membrane/rtp/session_bin.ex
```diff
@@ -361,23 +381,27 @@ defmodule Membrane.RTP.SessionBin do
...
       |> to(rtp_stream_name)
 
-    acc = {new_children, router_link}
+    acc = {new_children, router_link, []}
 
-    {new_children, router_link} =
+    {new_children, router_link, children_actions} =
       extensions
-      |> Enum.reduce(acc, fn {extension_name, config}, {new_children, new_link} ->
+      |> Enum.reduce(acc, fn {extension_name, header_extension_id, config},
```
Qizot, line 388: I don't think that those extensions always have something to do with the RTP extensions, it is a bad idea assume so.

Qizot (changes requested):

lib/membrane/rtp/session_bin.ex
balins: Maybe we could do something like this:
`extensions` -> `rtp_extensions`
`packet_filters` -> `extensions`
And use `packet_filters` as general extensions instead?

lib/membrane/rtp/outbound_packet_tracker.ex
Qizot: When I think about it, in the end there will be not many element so I don't know if it is worth the effort in the end, preassigning `Map.keys` is probably enough here 🚀

lib/membrane/rtp/session_bin.ex
Qizot: Yup, sounds great

balins: Renamed in 512ca6b

lib/membrane/rtp/outbound_packet_tracker.ex
balins: Guess we can resolve this ;)

Qizot: Yup, if you extract the `Map.keys` outside the filter :P

lib/membrane/rtp/vad.ex
```diff
@@ -108,8 +111,21 @@ defmodule Membrane.RTP.VAD do
 
   @impl true
   def handle_process(:input, %Membrane.Buffer{} = buffer, _ctx, state) do
-    vad_extension = Enum.find(buffer.metadata.rtp.extensions, &(&1.identifier == 1))
-    <<_v::1, level::7>> = vad_extension.data
+    {extension, buffer} = Header.Extension.pop_identifier(buffer, state.header_extension_id)
```
Qizot, line 114: My bad, when I look at it in this scenario `Header.Extension.pop` makes more sense than `pop_identifier`... Can you bring it back? 😢

lib/membrane/rtp/vad.ex
```diff
@@ -108,8 +111,21 @@ defmodule Membrane.RTP.VAD do
...
+    handle_if_present(buffer, extension, state)
+  end
+
+  defp handle_if_present(buffer, nil, state), do: {{:ok, buffer: {:output, buffer}}, state}
+
+  defp handle_if_present(buffer, extension, state) do
+    <<_v::1, level::7>> = extension.data
+
+    new_extension = %Header.Extension{
```
Qizot, line 123: Why do we pop the extension to bring it back again? You can add functions to just look up the extension if you don't intend to pop that.

lib/membrane/rtp/vad.ex
```diff
@@ -128,6 +144,11 @@ defmodule Membrane.RTP.VAD do
     end
   end
 
+  @impl true
+  def handle_other({:header_extension_id, id}, _ctx, state) do
+    {:ok, %{state | header_extension_id: id}}
```
Qizot, line 149: How about `vad_id` instead of `header_extension_id` in the state?

Qizot (changes requested):

lib/membrane/rtp/vad.ex
balins: Mostly because you have to put it back with locally-recognized identifier, ex. `:vad` instead of `1`. So even if there will be a function like `Header.Extension.present?`, you will have to check for extension's presence, then pop the extension with integer identifier and put one with locally-recognized one.
So instead of one pass through extension list (by `Header.Extension.pop`) you have one pass for checking for presence and second for popping/updating extension to provide a locally-recognized id.

lib/membrane/rtp/vad.ex
balins: Sounds great!

lib/membrane/rtp/vad.ex
balins: No worries, it's back

lib/membrane/rtp/outbound_packet_tracker.ex
balins: Ah, sure 😝

lib/membrane/rtp/vad.ex
Qizot: I'm wondering if we will ever need the identifier to be changed to `:vad`. Is any other element supposed to read that (for serialization or something)? I guess it is fine for debugging purposes but I am probably missing the use case for that in the long run.

balins: `OutboundPacketTracer` can remap these local identifiers to integer identifiers expected by a stream receiver.
So one peer, say Chrome, can send VAD with `extension_id=1` but we cannot forward this to other peers carelessly, because e.g. Safari client expects VAD under `extension_id=14`.  These extension ids vary among clients, and we are able to remap them to ids expected by clients only when we know who exactly the receiver is, that is not sooner than in `StreamSendBin`.

Qizot (approved):

balins:
I've replaced `forward` with passing extension ID on `rtp_extensions` initialization, as it seems less complicated. Could you take a look @Qizot?

PR #87, opened 2021-12-10
Update SilenceDiscarder

balins:
- adjusted `RTP.SilenceDiscarder` to the recent changes in RTP header extensions
- added `RTP.Header.Extension.find/2`
- updated moduledocs mentioning header extensions
- changed defaults in `RTP.SilenceDiscarder` options

Default for `silence_threshold` has been changed to quite arbitrary `64`, as this value seems to do pretty well. Previously set `127` was too strict to ever drop any packet 🤔

### DIFF
--- a/lib/membrane/rtp/silence_discarder.ex
+++ b/lib/membrane/rtp/silence_discarder.ex
@@ -2,7 +2,7 @@ defmodule Membrane.RTP.SilenceDiscarder do
   @moduledoc """
   Element responsible for dropping silent audio packets.
 
-  For a packet to be discarded it needs to contain a `RTP.HeaderExtension` struct in its
+  For a packet to be discarded it needs to contain a `RTP.Header.Extension` struct in its
   metadata under `:rtp` key. The header should contain information about audio level (VAD extension is required).
   The element will only drop packets whose audio level is above given silence threshold (muted audio is of value 127).
 
@@ -12,12 +12,7 @@ defmodule Membrane.RTP.SilenceDiscarder do
   """
   use Membrane.Filter
 
-  alias Membrane.RTP.Header
-  alias Membrane.RTP.PacketsDiscardedEvent
-
-  require Membrane.Logger
-
-  @vad_len 0
+  alias Membrane.RTP.{Header, PacketsDiscardedEvent}
 
   def_input_pad :input, caps: :any, demand_unit: :buffers
   def_output_pad :output, caps: :any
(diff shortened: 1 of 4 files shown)


lib/membrane/rtp/silence_discarder.ex
```diff
@@ -35,15 +30,15 @@ defmodule Membrane.RTP.SilenceDiscarder do
               ],
               silence_threshold: [
                 spec: 1..127,
-                default: 127,
+                default: 64,
```
Qizot, line 33: Why did you change that to 64? It can drop the normal speech then, 127 means silent.

lib/membrane/rtp/silence_discarder.ex
```diff
@@ -80,30 +70,26 @@ defmodule Membrane.RTP.SilenceDiscarder do
...
-          dropped > 0 ->
-            stop_dropping(buffer, state)
-
-          true ->
-            {{:ok, buffer: {:output, buffer}}, state}
-        end
-
-      _header ->
+    maybe_vad = Header.Extension.find(buffer, state.vad_id)
+    handle_vad(buffer, maybe_vad, state)
```
Qizot, line 74:
```suggestion
    buffer
    |> Header.Extension.find(state.vad_id)
    |> handle_vad(buffer, state)
```

lib/membrane/rtp/silence_discarder.ex
```diff
@@ -80,30 +70,26 @@ defmodule Membrane.RTP.SilenceDiscarder do
...
+  end
+
+  defp handle_vad(buffer, nil, state), do: {{:ok, buffer: {:output, buffer}}, state}
+
+  defp handle_vad(buffer, vad, state) do
+    %{dropped: dropped, silence_threshold: silence_threshold} = state
+    <<_v::1, audio_level::7>> = vad.data
+
+    silent? = audio_level >= silence_threshold
```
Qizot, line 83: Put that directly into the `cond`

Qizot (changes requested):

lib/membrane/rtp/silence_discarder.ex
balins: I explained in PR description, 127 would not drop anything in normal cases as there is always some ambient noise. Do we want to target only explicitly muted streams? If so, we can change it back. If we want to also consider some amount of ambient noise, we can think about a more safe default value.

Qizot: I think that for now we are good with dropping just the ambient noise (unless you want to spend more time experimenting with what audio level can be considered silence). The main case is when there are a lot of people in the room and most of them are muted but the packets still get sent, it is pretty much the main reason for this filter to exist.

Qizot (approved):

PR #88, opened 2022-01-20
Outbound TWCC

balins:
This PR includes:
- `RTP.TWCCSender` element for tagging outgoing packets with transport-wide sequence numbers
- decoding incoming TWCC feedback packets
- slightly improved encoding for TWCC feedbacks (from [`libwebrtc` implementation](https://github.com/mozilla/libwebrtc/blob/master/modules/rtp_rtcp/source/rtcp_packet/transport_feedback.cc), the last status vector in a feedback packet can be incomplete – there is no need to convert it to series of run length chunks)
- tests for encoding/decoding TWCC feedbacks


lib/membrane/rtcp/transport_feedback_packet/twcc.ex
```diff
@@ -89,13 +102,107 @@ defmodule Membrane.RTCP.TransportFeedbackPacket.TWCC do
...
+  end
+
+  defp parse_feedback(payload, packet_status_count) do
+    with {:ok, {encoded_receive_deltas, parsed_packet_status}} <-
+           parse_packet_status(payload, packet_status_count, []),
+         {:ok, receive_deltas} <-
+           parse_receive_deltas(encoded_receive_deltas, parsed_packet_status, []) do
+      {:ok, receive_deltas}
+    else
+      {:error, reason} -> {:error, reason}
```
Qizot, line 125: Do you need to explicitly match on `{:error, reason}`? You could just omit the `else`

test/membrane/rtcp/twcc_test.exs
```diff
@@ -0,0 +1,28 @@
...
+
+  describe "TWCC module" do
+    test "encodes and decodes valid feedbacks" do
+      encoded_feedbacks = Fixtures.twcc_feedbacks()
+      expected_feedbacks = Fixtures.twcc_feedbacks_contents()
+
+      encoded_feedbacks
+      |> Enum.zip(expected_feedbacks)
+      |> Enum.each(fn {encoded, expected} ->
+        assert TWCC.decode(encoded) == {:ok, expected}
```
Qizot, line 15: Try to put the expected value on the left side. 
`assert {:ok, expected} == TWCC.decode(encoded)`

lib/membrane/rtp/session_bin.ex
```diff
@@ -634,4 +638,27 @@ defmodule Membrane.RTP.SessionBin do
...
+
+  defp maybe_handle_twcc_sender(pad_ssrc, ctx) do
+    # Workaround: as TWCC is a transport-wide extension, there should exist only one TWCC sender
+    # child that handles packets for all outgoing streams. As there is no support for declaring
+    # outbound extensions, outbound TWCC will be enabled if inbound TWCC has been enabled.
+    should_link? = Map.has_key?(ctx.children, :twcc_receiver)
+    should_create_child? = not Map.has_key?(ctx.children, :twcc_sender)
+
+    if should_link? do
+      &(&1
```
Qizot, line 650: can you change it to `fn` in this if path? Would be more readable than it is now.

lib/membrane/rtcp/transport_feedback_packet/twcc.ex
```diff
@@ -89,13 +102,107 @@ defmodule Membrane.RTCP.TransportFeedbackPacket.TWCC do
...
+    end
+  end
+
+  defp parse_packet_status(binary, packets_left, parsed_status) when packets_left <= 0 do
```
Qizot, line 129: I would stay away from naming things plain `binary`.

Qizot (changes requested):

Qizot (approved):
LGTM

PR #89, opened 2022-02-04
switch to auto demands

mat-hek:


lib/membrane/rtp/outbound_packet_tracker.ex
```diff
@@ -138,21 +125,20 @@ defmodule Membrane.RTP.OutboundPacketTracker do
...
-
-    actions =
-      %{state.ssrc => stats}
-      |> SenderReport.generate_report()
-      |> Enum.map(&Membrane.RTCP.Packet.serialize(&1))
-      |> Enum.map(&{:buffer, {state.rtcp_output_pad, %Membrane.Buffer{payload: &1}}})
-
-    {{:ok, actions ++ [redemand: state.rtcp_output_pad]}, %{state | any_buffer_sent?: false}}
+  def handle_other(:send_stats, ctx, state) do
+    if state.rtcp_output_pad && not ctx.pads[state.rtcp_output_pad].end_of_stream? do
```
Qizot, line 129: nitpick: what about assigning `state.rtcp_output_pad` to some variable that is shorter? We use it 3 times and it is quite lengthy

lib/membrane/srtp/encryptor.ex
```diff
@@ -63,7 +64,9 @@ defmodule Membrane.SRTP.Encryptor do
     }
 
     :ok = ExLibSRTP.add_stream(state.srtp, policy)
-    {{:ok, redemand: :output}, Map.put(state, :policies, [policy])}
+
+    {{:ok, buffer: {:output, Enum.reverse(state.queue)}},
```
Qizot, line 68: What is the purpose of the queue?

If there are no policies we would end up caching the buffers and when we get the handshake we just send them all without encryption?

Qizot (changes requested):

lib/membrane/srtp/encryptor.ex
mat-hek: The problem is that SRTP policy may come via output pad and till now we could hold demands until we receive it, which is not possible with auto demands, so we have to buffer them and encrypt them once we get the policy.

> we just send them all without encryption

good point, fixed :P

Qizot (approved):

PR #91, opened 2022-02-09
Fixing TWCC Receiver executing actions on unlinked pads

daniel-jodlos:


lib/membrane/rtp/twcc_receiver.ex
```diff
@@ -35,6 +35,7 @@ defmodule Membrane.RTP.TWCCReceiver do
 
   defmodule State do
     @moduledoc false
+    use Bunch.Access
```
balins, line 38: Do we use it anywhere?

lib/membrane/rtp/twcc_receiver.ex
```diff
@@ -150,4 +159,30 @@ defmodule Membrane.RTP.TWCCReceiver do
       }
     }
   end
+
+  # Checks if actions would be valid in the current context, namely, if they don't reference the pad that hasn't been linked yet
+  # If they are valid, they are executed immediately, otherwise they are buffered and executed as soon as the pad in question is added
+  defp send_as_actions_or_buffer(actions, ctx, state) do
```
balins, line 165: It's a nice, generalized solution to the problem, but I'm worried if it's an overkill for this use case. Callbacks in this element send actions on only one pad at a time, so maybe we could just check if `Map.has_key?(ctx.pads, the_pad)` and then store or send immediately all at once?

lib/membrane/rtp/twcc_receiver.ex
```diff
@@ -150,4 +159,30 @@ defmodule Membrane.RTP.TWCCReceiver do
...
+        {:end_of_stream, pad} ->
+          is_valid?.(pad)
+
+        _otherwise ->
+          :valid
+      end)
+
+    buffer =
+      Map.get(grouped, :invalid, [])
+      |> Enum.reduce(state.actions_buffer, fn {_action, {pad, _args}} = action, buffer ->
```
balins, line 182: This will crash on `{:end_of_stream, pad} = action`

lib/membrane/rtp/twcc_receiver.ex
```diff
@@ -71,14 +74,20 @@ defmodule Membrane.RTP.TWCCReceiver do
...
-    {{:ok, caps: {Pad.ref(:output, ssrc), caps}}, state}
+  def handle_caps(Pad.ref(:input, ssrc), caps, ctx, state) do
+    [caps: {Pad.ref(:output, ssrc), caps}] |> send_as_actions_or_buffer(ctx, state)
   end
 
   @impl true
-  def handle_event(Pad.ref(direction, ssrc), event, _ctx, state) do
+  def handle_pad_added(pad, _ctx, state) do
+    {actions, buffer} = Map.pop(state.actions_buffer, pad, [])
+    {{:ok, Enum.to_list(actions)}, %{state | actions_buffer: buffer}}
```
balins, line 84: Maybe we could rename `actions_buffer` to `buffered_actions`? When reading `buffer` in a callback, the  `%Membrane.Buffer{}` is always the first thing that comes to my mind 😅

balins (changes requested):

lib/membrane/rtp/twcc_receiver.ex
daniel-jodlos: Fine, but to minimize potential for nasty bugs in the future, the if statement will have to appear in every single callback. Otherwise it would be asking for trouble if this property ever changes

balins: I was thinking about passing `Map.has_key?(ctx.pads, the_pad)` as the function parameter, and then matching in the `send_as_actions_or_buffer` function header on `true`/`false`. This way it will be explicit that with this implementation we are using only one output pad at a time, and we can reduce the performance impact to the necessary minimum.

daniel-jodlos: That sounds weird to me. I think I'd prefer having the `if` statement. - that would be even more explicit
```elixir
if Map.has_key?(ctx.pads, pad), do: {{:ok, actions}, state}, else: {:ok, buffer(actions, state)}
```

balins: I wanted to keep the logic related to pad presence separate, but if you prefer having it in callbacks, we can go with that.

lib/membrane/rtp/twcc_receiver.ex
daniel-jodlos: Yes

lib/membrane/rtp/twcc_receiver.ex
daniel-jodlos: This is no longer relevant

lib/membrane/rtp/twcc_receiver.ex
balins: Double checked, there is no place where we use it

balins (approved):

PR #95, opened 2022-02-22
don't use deprecated logger

mat-hek:


mix.exs
```diff
@@ -73,10 +73,9 @@ defmodule Membrane.RTP.Plugin.MixProject do
 
   defp deps do
     [
-      {:membrane_core,
-       github: "membraneframework/membrane_core", tag: "v0.9.0-rc.0", override: true},
+      {:membrane_core, "~> 0.9.0", override: true},
```
bblaszkow06, line 76: Why override? You should rather update elements. `h264_ffmpeg` has already 0.17 ready for core 0.9

mix.exs
```diff
@@ -73,10 +73,9 @@ defmodule Membrane.RTP.Plugin.MixProject do
...
       {:membrane_rtp_format, "~> 0.3.1"},
-      {:membrane_rtp_h264_plugin, "~> 0.8.0"},
+      {:membrane_rtp_h264_plugin, "~> 0.8.0", only: :test},
```
bblaszkow06, line 78: I usually removed the last number from test only deps as they won't be published and locally `mix.lock` takes care of ensuring the proper version - here it would allow updating h264_ffmpeg to version compatible with new core

lib/membrane/rtp/jitter_buffer.ex
```diff
@@ -90,7 +90,7 @@ defmodule Membrane.RTP.JitterBuffer do
           %State{state | store: result}
 
         {:error, :late_packet} ->
-          warn("Late packet has arrived")
+          Membrane.Logger.warn("Late packet has arrived")
```
bblaszkow06, line 93: image

bblaszkow06 (changes requested):

mix.lock
```diff
@@ -26,24 +26,26 @@
...
   "membrane_core": {:hex, :membrane_core, "0.9.0", "91a79636c6b2fcc6ae2e611670cdca33a0b38ff4dabe8fb7c1d56cba426db787", [:mix], [{:bunch, "~> 1.3", [hex: :bunch, repo: "hexpm", optional: false]}, {:qex, "~> 0.3", [hex: :qex, repo: "hexpm", optional: false]}, {:ratio, "~> 2.0", [hex: :ratio, repo: "hexpm", optional: false]}, {:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "ab252e7aa944d23fbb22ce0e8bc75b088cfe3ed1c2c7ba203b1cc2fd794a85ed"},
   "membrane_element_pcap": {:git, "https://github.com/membraneframework/membrane-element-pcap.git", "e9d9c47bc1ed5897250e6463710b4adb0d76c4ed", [tag: "v0.4.0"]},
   "membrane_element_udp": {:hex, :membrane_element_udp, "0.6.0", "40174620e1e08bb5e0dc498f5b2aec16eb29b6995127e43d1bd45599147d5d9f", [:mix], [{:membrane_core, "~> 0.8.0", [hex: :membrane_core, repo: "hexpm", optional: false]}, {:mockery, "~> 2.3.0", [hex: :mockery, repo: "hexpm", optional: false]}], "hexpm", "ce7d9941ec65628bba7e8da4d7414cd588fc40c794a2a609859c85cdfcc718e2"},
```
bblaszkow06, line 32: Please run `mix deps.unlock --unused`

mix.exs
```diff
@@ -73,25 +73,25 @@ defmodule Membrane.RTP.Plugin.MixProject do
...
+      {:membrane_rtp_mpegaudio_plugin, "~> 0.8.0", only: :test},
+      {:membrane_h264_ffmpeg_plugin, "~> 0.17.0", only: :test},
+      {:membrane_pcap_plugin,
+       github: "membraneframework/membrane_pcap_plugin", tag: "v0.5.0", only: :test},
+      {:membrane_hackney_plugin, "~> 0.7.0", only: :test},
+
       # Dev
       {:ex_doc, "~> 0.21", only: :dev, runtime: false},
       {:dialyxir, "~> 1.0", only: :dev, runtime: false},
-      {:excoveralls, ">= 0.8.0", only: :test},
```
bblaszkow06, line 95: If you're removing this, remove the leftovers in `project` callback

bblaszkow06: you can also apply the usual stuff to `docs`, i.e. `LICENSE: [title: License]` and `formatters: ["HTML"],`

bblaszkow06 (approved):
I don't think I have to re-review after fixes, so :heavy_check_mark:

PR #99, opened 2022-04-14
Change warn logs to debug

mickel8:


lib/membrane/srtp/encryptor.ex
```diff
@@ -90,7 +90,7 @@ if Code.ensure_loaded?(ExLibSRTP) do
 
     @impl true
     def handle_event(_pad, other, _ctx, state) do
-      Membrane.Logger.warn("Got unexpected event: #{inspect(other)}. Ignoring.")
+      Membrane.Logger.debug("Got unexpected event: #{inspect(other)}. Ignoring.")
```
mickel8, line 93: I am not sure about this one :thinking:

mat-hek: AFAIR we discussed that and decided to start with checking what are the packets triggering this

mickel8: Do you mean log that warns that we couldn't parse RTCP packet? This one warns about unexpected `Membrane.Event`. Anyway, there are still logs like `Received PLI` or logs related to TWCC that we should probably log on debug only

mat-hek: Ok, I see that now. This is quite weird. I think it should be like this:
- This clause should match on the handshake event only
- There should be another clause, matching all other events, with the default behaviour (calling `super`)
- This clause should explicitly match on the handshake event as well https://github.com/membraneframework/membrane_rtp_plugin/pull/99/files/321d90ece8be92a25dc3c2244984625f9c8e2b90#diff-b36b4c392e517014f0a9c2962af852149ee1c3a9ab3d504d22fcfdf3c8101ed5R73

mickel8: I have fixed that, however the link doesn't redirect me to any specific module

FelonEkonom (approved):

PR #100, opened 2022-05-04
Release membrane_rtp_plugin v0.12.0

varsill:
### Release "Vastly Pro Piglet"

| Project | New Version |
|---|---|
| membrane_core | `0.10.0` |

<sup>proudly automated with <a href="https://github.com/membraneframework/sebex">Sebex</a></sup>


lib/membrane/rtcp/parser.ex
```diff
@@ -24,7 +24,13 @@ defmodule Membrane.RTCP.Parser do
 
   @impl true
   def handle_caps(:input, _caps, _ctx, state) do
-    {{:ok, caps: {:output, %RTCP{}}}, state}
```
mickel8, line 27: Caps are needed only when we want to send buffers and we don't send any buffers over `:output` pad. We send only events. When I moved sending these caps to `handle_prepared_to_playing`, SSRCRouter crashed as it is not prepared for receiving RTCP caps, only RTP ones. I think that `handle_caps` callback has been never called in this element. Not sure what we want to do with this? IMO, we should fix SSRCRouter but I would move this to a separate PR to not block releasing other packages.

cc @mat-hek

mickel8:
I have fixed most tests however there are still two that are not passing. I couldn't also get rid of `Testing.Pipeline.play` or I didn't know how to do this :thinking:

lib/membrane/rtcp/parser.ex
mat-hek: I think caps may be useful no matter if we send buffers or not. Maybe one day we should enforce sending them before sending any events as well? I guess fixing SSRCRouter is the proper way for this case 👍

lib/membrane/rtcp/parser.ex
```diff
@@ -15,13 +15,23 @@ defmodule Membrane.RTCP.Parser do
     demand_mode: :auto
 
   def_output_pad :output, caps: RTCP, demand_mode: :auto
-  def_output_pad :rtcp_output, mode: :push, caps: :any
+
+  def_output_pad :receiver_report_output,
+    mode: :push,
+    caps: {RemoteStream, type: :packetized, content_format: one_of([nil, RTCP])}
```
mickel8, line 21: As we disscussed, I would write just `RTCP`.

mickel8 (changes requested):

README.md
```diff
@@ -14,7 +14,7 @@ The package can be installed by adding `membrane_rtp_plugin` to your list of dep
 ```elixir
 def deps do
   [
-    {:membrane_rtp_plugin, "~> 0.11.0"}
+	{:membrane_rtp_plugin, "~> 0.12.0"}
```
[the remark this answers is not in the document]
varsill: sure ;)

lib/membrane/rtcp/parser.ex
varsill: done

mickel8 (approved):

PR #101, opened 2022-05-09
Fix forwarding :end_of_stream in SSRC Router. Refactor SSRC Router

mickel8:
Sometimes, when the input is pretty small and we use auto demands, we might get `:end_of_stream` on our input pad before output pad for given ssrc is linked (auto-demands loads the whole input at start). In such a case we will discard `:end_of_stream` action and we will not forward it after output pad is linked. This PR buffers `:end_of_stream` together with other actions to forward it when an output pad for given ssrc appears.

### DIFF
--- a/lib/membrane/rtp/ssrc_router.ex
+++ b/lib/membrane/rtp/ssrc_router.ex
@@ -56,7 +56,8 @@ defmodule Membrane.RTP.SSRCRouter do
       state.input_pads
       |> Enum.filter(fn {_ssrc, p} -> p == pad end)
       |> Enum.flat_map_reduce(state, fn {ssrc, _pad}, state ->
-        maybe_buffer_action(:end_of_stream, ssrc, ctx, state)
+        action = {:end_of_stream, Pad.ref(:output, ssrc)}
+        maybe_buffer_action(action, ssrc, ctx, state)
       end)
 
     {{:ok, actions}, state}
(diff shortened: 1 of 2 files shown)


mickel8:
I think I should write tests for this module but I really want to realse RTC Engine so maybe I could do this in a separate PR? Or better do this now?

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -157,22 +162,30 @@ defmodule Membrane.RTP.SSRCRouter do
...
-  defp maybe_add_to_linking_buffer(type, value, ssrc, state) do
+  defp maybe_buffer_action(:end_of_stream, ssrc, ctx, state) do
+    action = {:end_of_stream, Pad.ref(:output, ssrc)}
+    do_maybe_buffer_action(action, ssrc, ctx, state)
+  end
+
+  defp maybe_buffer_action(type, value, ssrc, ctx, state) do
     action = {type, {Pad.ref(:output, ssrc), value}}
+    do_maybe_buffer_action(action, ssrc, ctx, state)
+  end
```
varsill, line 179: Hmm I am not sure if we should distinguish between different `Membrane.Element.Action.t` types in function signature...I know it is just a private function, but personally I find it less readable. How about removing these two "converter" functions (`maybe_buffer_action/4` and maybe_buffer_action/5`) and explicitly calling `do_maybe_buffer_actions/4` (obviously renamed to `maybe_buffer_actions/4` with the `Membrane.Element.Action.t` as first argument?

varsill (commented):
I don't have any objections concerning such a action execution postoping mechanism. There is just a small comment on code readability, but it might be just my personal feeling ;D

varsill (approved):

mat-hek (approved):

PR #104, opened 2022-05-17
Support telemetry

FelonEkonom:


lib/membrane/rtp/ssrc_router.ex
```diff
@@ -15,9 +15,32 @@ defmodule Membrane.RTP.SSRCRouter do
...
+    demand_mode: :auto,
+    options: [
+      keyframe_detector: [
+        spec: function() | nil,
+        default: nil
+      ],
+      frame_detector: [
+        spec: function() | nil,
+        default: nil
+      ],
```
mickel8, line 34: I would encapsulate this in one function called e.g. `stats`.

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -179,5 +227,64 @@ defmodule Membrane.RTP.SSRCRouter do
...
+      telemetry_event_name(),
+      ctx.pads[pad].options.telemetry_metadata
+    )
+  end
+
+  defp emit_packet_arrival_event(ssrc, payload, pad, state, ctx) do
+    Membrane.TelemetryMetrics.execute(
+      telemetry_event_name(),
+      packet_arrival_telemetry_measurements(ssrc, payload, pad, state, ctx),
+      %{ssrc: ssrc, telemetry_metadata: ctx.pads[pad].options.telemetry_metadata}
```
mat-hek: Why do we have a `telemetry_metadata` field inside, well, telemetry metadata? :P And why is `ssrc` separate?

lib/membrane/rtp/ssrc_router.ex
mat-hek: Maybe ssrc router could take these himself using `PayloadFormat.get`? Also, maybe create a behaviour and have a module containing `detect_keyframe` and `detect_frame` for each protocol?

mat-hek (changes requested):

lib/membrane/rtp/ssrc_router.ex
FelonEkonom: There are 2 issues with this approach:
1. Only VP8 format has both `frame_detector` and `keyframe_detector`, OPUS and H264 have only one of these
2. Currently, there is no dependency in `membrane_rtp_plugin` to `membrane_rtp_vp8_plugin` etc, so I would have to add them or pass this module in place, where I pass `frame_detector` and `keyframe_detector`, so I don't think if doing this would change a lot (assuming, that I don't want to add these dependencies)

mat-hek: SSRC router could get the module the way Session Bin gets payloaders and depayloaders. That won't require RTP plugin to depend on rtp_vp8 etc, just like it doesn't depend on them now, still being able to get payloaders. Only the end-user would have to depend on them for this to work. If some codecs don't implement particular functions, they could either not have them in the said module and we'd have to check for them, or we could provide some default if possible.

lib/membrane/rtcp/receiver.ex
```diff
@@ -20,12 +20,13 @@ defmodule Membrane.RTCP.Receiver do
               remote_ssrc: [spec: RTP.ssrc_t()],
               report_interval: [spec: Membrane.Time.t() | nil, default: nil],
               fir_interval: [spec: Membrane.Time.t() | nil, default: nil],
-              telemetry_metadata: [spec: [{atom(), any()}], default: []]
+              telemetry_label: [spec: Membrane.TelemetryMetrics.label(), default: []]
 
-  @event_name [:sending_fir, :rtcp]
+  @event_name [:RTCP, :fir, :sending]
```
mickel8, line 25: Wouldn't it be better to name this `sent` instead of `sending`?

```suggestion
  @event_name [:RTCP, :fir, :sent]
```

mickel8 (approved):

lib/membrane/rtp/metrics.ex
```diff
@@ -1,30 +1,35 @@
...
     [
       Telemetry.Metrics.counter(
         "inbound-rtp.keyframe_request_sent",
-        event_name: [:sending_fir, :rtcp]
+        event_name: [Membrane.RTP, :RTCP, :fir, :sending]
       ),
       Telemetry.Metrics.counter(
         "inbound-rtp.packets",
-        event_name: [:packet_arrival, :rtp]
+        event_name: [Membrane.RTP, :RTP, :packet, :arrival]
```
mat-hek, line 18: I'd prefix RTCP events only, or at least not use capital letters

```suggestion
        event_name: [Membrane.RTP, :rtp, :packet, :arrival]
```

lib/membrane/rtp/payload_type_resolver.ex
```diff
@@ -31,13 +31,15 @@ defmodule Membrane.RTP.PayloadFormatResolver do
     end
   end
 
+  @spec keyframe_detector(atom()) :: {:ok, function()} | :error
   def keyframe_detector(encoding) do
     case PayloadFormat.get(encoding).keyframe_detector do
       nil -> :error
       keyframe_detector -> {:ok, keyframe_detector}
     end
   end
```
mat-hek, line 40: I guess this is unused and to be refactored, so I'd remove it

mat-hek (changes requested):

lib/membrane/rtp/payload_type_resolver.ex
FelonEkonom: It is used in RTC Engine

lib/membrane/rtp/payload_type_resolver.ex
```diff
@@ -31,13 +31,15 @@ defmodule Membrane.RTP.PayloadFormatResolver do
...
 
+  @spec frame_detector(atom()) :: {:ok, function()} | :error
   def frame_detector(encoding) do
     case PayloadFormat.get(encoding).frame_detector do
```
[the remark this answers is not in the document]
FelonEkonom: It is used in RTC Engine

mat-hek (approved):

PR #106, opened 2022-05-23
Fix determining late packets for big JitterBuffer latencies.

mickel8:
Sometimes, when the `latency` in `JitterBuffer` is big enough (e.g. 5 seconds), we might start discarding packets treating them as late. Included test presents this scenario. I think that there is also a `TODO` tag directly related to this problem https://github.com/membraneframework/membrane_rtp_plugin/blob/b05114afe0ac7dacefb9488e9fb4190892b82519/lib/membrane/rtp/jitter_buffer/buffer_store.ex#L86

The initial proposal was to use `end_index` in `BufferStore` for determining late packets. However, I encountered following problems
1. This messes up current architecture of `BufferStore` and `JitterBuffer`. At the moment, we are increasing ROC (rollover count) when performing `shift` operation. Using `end_index` requires handling rollovers when adding a new buffer.
2. I am not sure what is initial latency used for. I feel like it could be deleted.
3. I feel like our JitterBuffer does not compensate network jitter. It only reorders packets. 

Having said that, I think that the decision whether we want to fix this test or not and how to do this requires further discussion.

You can run failing test with `mix test --only debug`

cc @mat-hek @bblaszkow06

### DIFF
--- a/lib/membrane/rtp/jitter_buffer/buffer_store.ex
+++ b/lib/membrane/rtp/jitter_buffer/buffer_store.ex
@@ -14,7 +14,7 @@ defmodule Membrane.RTP.JitterBuffer.BufferStore do
 
   @seq_number_limit Bitwise.bsl(1, 16)
 
-  defstruct shift_index: nil,
+  defstruct flush_index: nil,
             highest_incoming_index: nil,
             heap: Heap.new(&Record.rtp_comparator/2),
             set: MapSet.new(),
@@ -27,13 +27,13 @@ defmodule Membrane.RTP.JitterBuffer.BufferStore do
   - `rollover_count` - count of all performed rollovers (cycles of sequence number)
   - `heap` - contains records containing buffers
   - `set` - helper structure for faster read operations; content is the same as in `heap`
-  - `shift_index` - index of the last packet that has been emitted as a result of a call
-  to one of the `shift` functions
+  - `flush_index` - index of the last packet that has been emitted as a result of a call
+  to one of the `flush` functions
   - `highest_incoming_index` - the highest index in the buffer so far, mapping to the most recently produced
   RTP packet placed in JitterBuffer
   """
   @type t :: %__MODULE__{
-          shift_index: JitterBuffer.packet_index() | nil,
+          flush_index: JitterBuffer.packet_index() | nil,
           highest_incoming_index: JitterBuffer.packet_index() | nil,
           heap: Heap.t(),
           set: MapSet.t(),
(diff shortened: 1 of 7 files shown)


lib/membrane/rtp/jitter_buffer/buffer_store.ex
```diff
@@ -211,19 +210,16 @@ defmodule Membrane.RTP.JitterBuffer.BufferStore do
...
       |> update_end_index(record.index)
+      |> update_roc(record_epoch)
     end
   end
 
-  defp bump_prev_index(%{prev_index: prev, rollover_count: roc} = store)
-       when rem(prev + 1, @seq_number_limit) == 0,
-       do: %__MODULE__{store | prev_index: prev + 1, rollover_count: roc + 1}
-
   defp bump_prev_index(store), do: %__MODULE__{store | prev_index: store.prev_index + 1}
```
mat-hek, line 223: - I think it doesn't deserve a separate function now, or at least give it `update` prefix as the others have
- Let's rename this `prev_index` to something more meaningful, or at least less misleading :P Maybe `start_index`?

mat-hek (commented):
LGTM, I'd only suggest some small refactor

- Maybe rename `rollover_count` to `epoch` to use one word?
- I'd move the `fields` section of the comment at the top to the typedoc for `t` type
- The function `size` doesn't calculate the size, as it explicitly states in its doc :P Maybe something like `range` would be better?

I assume relying on packets' timestamps will be in another PR?

mickel8:
> Maybe rename rollover_count to epoch to use one word?

ROC is commonly used term in RFCs so I would keep it.

My final proposal is as follows:
* `shift` functions -> `flush` functions
* `prev_index` -> `flush_index`
* `end_index` -> `highest_incoming_index`

mickel8:
> I assume relying on packets' timestamps will be in another PR?

After this fix, I don't fully get how we could benefit from relying on RTP timestamps :thinking: Could we discuss this internally, besides this PR?

lib/membrane/rtp/jitter_buffer/buffer_store.ex
```diff
@@ -21,15 +14,27 @@ defmodule Membrane.RTP.JitterBuffer.BufferStore do
...
             rollover_count: 0
 
+  @typedoc """
+  Type describing BufferStore structure.
+
+  Fields:
+  - `rollover_count` - count of all performed rollovers (cycles of sequence number)
+  - `heap` - contains records containing buffers
+  - `set` - helper structure for faster read operations; content is the same as in `heap`
+  - `flush_index` - index of the last packet that has been emitted as a result of a call
```
mat-hek, line 30:
```suggestion
  - `flush_index` - index of the last packet that has been emitted (or would have been emitted, but never arrived) as a result of a call
```

mat-hek (commented):
If we want to stick to `roc` I'd rename `epoch` to `rollover`

mickel8:
Honestly, I don't feel the last commit but if you think that it's good enough we can keep it

PR #107, opened 2022-07-04
Feat/fir pli handling

bblaszkow06:


lib/membrane/rtp/outbound_packet_tracker.ex
```diff
@@ -91,6 +94,21 @@ defmodule Membrane.RTP.OutboundPacketTracker do
...
+  @impl true
+  def handle_event(
+        Pad.ref(:rtcp_input, _id),
+        %RTCPEvent{rtcp: %{payload: %keyframe_request{}}},
+        _ctx,
+        state
+      )
+      when keyframe_request in [PLI, FIR] do
+    {{:ok, event: {:input, %Membrane.KeyframeRequestEvent{}}}, state}
+  end
```
mickel8, line 106: A word of comment here would be appreciated

config/config.exs
```diff
@@ -0,0 +1,3 @@
+import Config
```
mickel8, line 1: Is this needed? We have configured default in code to 500 (RTCP Receiver l. 27). If we want to keep this I would use `Application.compile_env!` in RTCP Receiver.

test/membrane/rtp/stream_receive_bin_test.exs
```diff
@@ -102,4 +105,84 @@ defmodule Membrane.RTP.StreamReceiveBinTest do
     assert_end_of_stream(pipeline, :sink)
     Testing.Pipeline.terminate(pipeline, blocking?: true)
   end
+
+  defmodule NoopSource do
```
mickel8, line 109: Sounds like something that should be achievable with TestingSource :thinking:

test/membrane/rtp/stream_receive_bin_test.exs
```diff
@@ -102,4 +105,84 @@ defmodule Membrane.RTP.StreamReceiveBinTest do
...
+      {{:ok, [keyframe_request, keyframe_request]}, delay}
+    end
+
+    @impl true
+    def handle_other(keyframe_request, _ctx, delay) do
+      {{:ok, [keyframe_request, keyframe_request]}, delay}
+    end
+  end
+
+  test "FIR sending with throttle" do
```
mickel8, line 145: in a lot of places we read tests as `test if x does something` but I am not sure if this the convention in rtp plugin

test/membrane/rtp/stream_receive_bin_test.exs
bblaszkow06: I thought that as well, the trouble is that it must send data and if passed an empty list, it sends EoS right away.
Maybe if we had `Testing.PushSource` it would be helpful?

config/config.exs
bblaszkow06: > If we want to keep this I would use Application.compile_env! in RTCP Receiver.

This is problematic as it would cause the compilation to fail in every packet that depends (even transitively) on it. In general, configs in libs should be avoided, thus originally I didn't have the config, just a module attribute. I've added it for testing purposes, as
1. I needed to access the value to not hardcode anything in tests
2. the test obviously time-sensitive and I expected issues with the CI, so I wanted to allow adjusting that
So you can consider it as a way of mocking the throttle for the test only

Do you think wrapping the config with an `if config_env == :test` would be better? Or adding a comment?

bblaszkow06: PS I'm open for any other suggestions as I'm not super happy with that config either

lib/membrane/rtcp/parser.ex
```diff
@@ -63,16 +63,14 @@ defmodule Membrane.RTCP.Parser do
     {{:ok, buffer: {:receiver_report_output, buffer}}, state}
   end
 
-  @impl true
-  def handle_event(pad, event, ctx, state), do: super(pad, event, ctx, state)
```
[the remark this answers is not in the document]
bblaszkow06: We would, but we don't expect anything else at the moment (or at least nothing appears here when used in videoroom). I think in general that we should prefer handling every event explicitly, but I guess this all has to be refactored anyway, so I'll bring it back

test/membrane/rtp/stream_receive_bin_test.exs
```diff
@@ -102,4 +105,84 @@ defmodule Membrane.RTP.StreamReceiveBinTest do
...
+
+    # Ensure we're not getting it twice before throttle duration passes
+    refute_pipeline_notified(pipeline, :src, %Membrane.RTCPEvent{}, half_throttle_duration)
+
+    # Then ensure we get the next one
+    assert_pipeline_notified(
+      pipeline,
+      :src,
+      %Membrane.RTCPEvent{rtcp: rtcp},
+      @fir_throttle_duration_ms
```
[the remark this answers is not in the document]
bblaszkow06: I'd like to avoid absolute values here, I'll use a value relative to `@fir_throttle_duration`

config/config.exs
mickel8: What about passing throttle_duration as an option to RTCP Receiver element?

test/membrane/rtp/stream_receive_bin_test.exs
mickel8: I see, I didn't think about EoS. I would keep it

mickel8 (approved):
There might be possible one change but it shouldn't be blocking

PR #109, opened 2022-08-03
[MV-116] Fix creating TWCC Sender

mickel8:


lib/membrane/rtp/session_bin.ex
```diff
@@ -66,13 +66,15 @@ defmodule Membrane.RTP.SessionBin do
...
-  ### Currently supported RTP extensions are:
+  ### RTP plugin ships with the following extensions:
   * `Membrane.RTP.VAD`
   * `Membrane.RTP.TWCCReceiver`
+  * `Membrane.RTP.TWCCSender`
 
   ### Example usage
   `{:vad, %Mebrane.RTP.VAD{vad_id: 1, time_window: 1_000_000}}`
   `{:twcc, %Mebrane.RTP.TWCCReceiver{twcc_id: 1, report_interval: Membrane.Time.milliseconds(250)}}`
+  `{:twcc, Mebrane.RTP.TWCCSender}`
```
bblaszkow06, line 77: This isn't rendered properly, I guess this should be a list
```suggestion
  * `{:vad, %Mebrane.RTP.VAD{vad_id: 1, time_window: 1_000_000}}`
  * `{:twcc, %Mebrane.RTP.TWCCReceiver{twcc_id: 1, report_interval: Membrane.Time.milliseconds(250)}}`
  * `{:twcc, Mebrane.RTP.TWCCSender}`
```

bblaszkow06: BTW, the typedoc doesn't seem like a right place for such docs - the examples should rather be in the pad options descriptions

lib/membrane/rtp/session_bin.ex
```diff
@@ -682,28 +696,33 @@ defmodule Membrane.RTP.SessionBin do
...
   end
 
-  defp maybe_handle_twcc_sender(pad_ssrc, ctx) do
+  defp maybe_handle_twcc_sender(rtp_extensions, pad_ssrc, ctx) do
     # Workaround: as TWCC is a transport-wide extension, there should exist only one TWCC sender
-    # child that handles packets for all outgoing streams. As there is no support for declaring
-    # outbound extensions, outbound TWCC will be enabled if inbound TWCC has been enabled.
-    should_link? = Map.has_key?(ctx.children, :twcc_receiver)
+    # child that handles packets for all outgoing streams.
+    {maybe_twcc, rtp_extensions} = Keyword.pop(rtp_extensions, :twcc)
```
bblaszkow06, line 702: There's no need to modify the `rtp_extensions`, they are ignored after return.
I'd rather use `get`
```suggestion
    maybe_twcc = Keyword.get(rtp_extensions, :twcc)
```

lib/membrane/rtp/session_bin.ex
```diff
@@ -682,28 +696,33 @@ defmodule Membrane.RTP.SessionBin do
...
+          |> via_out(Pad.ref(:output, pad_ssrc))
+        end
+      else
+        & &1
       end
-    else
-      & &1
-    end
+
+    {rtp_extensions, maybe_link_twcc}
```
bblaszkow06, line 726: With the change above you can return just the result of the `if`

lib/membrane/rtp/session_bin.ex
```diff
@@ -682,28 +696,33 @@ defmodule Membrane.RTP.SessionBin do
...
+              to(link, :twcc_sender, maybe_twcc)
+            else
+              to(link, :twcc_sender)
+            end
+          end)
+          |> via_out(Pad.ref(:output, pad_ssrc))
+        end
+      else
+        & &1
       end
```
bblaszkow06, line 724: I'd try to limit the nesting here
```suggestion
    link_twcc = fn link_builder ->
      link_builder
      |> via_in(Pad.ref(:input, pad_ssrc))
      |> then(fn link ->
        if should_create_child? do
          to(link, :twcc_sender, maybe_twcc)
        else
          to(link, :twcc_sender)
        end
      end)
      |> via_out(Pad.ref(:output, pad_ssrc))
    end

    if should_link? do
      link_twcc
    else
      & &1
    end
```

lib/membrane/rtp/session_bin.ex
```diff
@@ -66,13 +66,15 @@ defmodule Membrane.RTP.SessionBin do
 
   An extension will be spawned inside the bin under `{extension_name, ssrc}` name.
 
-  ### Currently supported RTP extensions are:
+  ### RTP plugin ships with the following extensions:
   * `Membrane.RTP.VAD`
   * `Membrane.RTP.TWCCReceiver`
+  * `Membrane.RTP.TWCCSender`
```
bblaszkow06, line 72: Just an idea - might be considered outside the scope of this PR - why do we add `:twcc` & `:vad` keys if we also use unique atoms as a value? It should be enough to pass just the module atoms/structs, e.g. instead of
```elixir
[
  vad: %Mebrane.RTP.VAD{vad_id: 1, time_window: 1_000_000},
  twcc: %Mebrane.RTP.TWCCReceiver{twcc_id: 1, report_interval: Membrane.Time.milliseconds(250)}
]
```
we could just do:
```elixir
[
  %Mebrane.RTP.VAD{vad_id: 1, time_window: 1_000_000},
  %Mebrane.RTP.TWCCReceiver{twcc_id: 1, report_interval: Membrane.Time.milliseconds(250)}
]
```

lib/membrane/rtp/session_bin.ex
```diff
@@ -219,10 +230,10 @@ defmodule Membrane.RTP.SessionBin do
         spec: [rtp_extension_options_t()],
         default: [],
         description: """
-        List of RTP extension options. Currently `:vad` and `:twcc` are supported.
+        List of RTP extension options. Currently `:vad` and `:twcc` (receiver) are supported.
```
bblaszkow06, line 233: This description is not enough to inform that only `{:twcc, %TWCCReceiver{}}` is allowed, IMO it should be more specific (or we can get rid of that key end enumerate just the modules :wink: )

bblaszkow06 (changes requested):

lib/membrane/rtp/session_bin.ex
mickel8: Looks like the `extension_name` is used for naming pipeline child corresponding to the given extension https://github.com/membraneframework/membrane_rtp_plugin/blob/master/lib/membrane/rtp/session_bin.ex#L434

and it is also allowed in `Membrane.RTP.Header.Extension.t`. I am not sure why :thinking:

lib/membrane/rtp/session_bin.ex
mickel8: This was prepared for future extensions but I can change it

lib/membrane/rtp/session_bin.ex
```diff
@@ -661,47 +682,52 @@ defmodule Membrane.RTP.SessionBin do
...
-          end
-        end)
-        |> via_out(Pad.ref(:output, pad_ssrc))
+    to_twcc_sender = fn link ->
+      if should_create_child? do
+        to(link, :twcc_sender, maybe_twcc)
+      else
+        to(link, :twcc_sender)
       end
+    end
```
mickel8, line 720: I had to move this out of the `link_twcc` as Credo was complaining about too deep nesting :thinking:

bblaszkow06 (approved):

PR #115, opened 2022-10-17
Fix selection of bwe increase mode. Fix bwe value when going into the decrease state.

mickel8:
* selection between additive and multiplicative mode is now based on r_hats from the decrease state
* when decreasing bandwidth estimation, `last_r_hat` is used instead of `last_a_hat` i.e. `est = 0.85 * last_r_hat`

### DIFF
--- a/lib/membrane/rtp/twcc_sender.ex
+++ b/lib/membrane/rtp/twcc_sender.ex
@@ -8,6 +8,7 @@ defmodule Membrane.RTP.TWCCSender do
   require Bitwise
 
   alias __MODULE__.CongestionControl
+  alias __MODULE__.ReceiverRate
   alias Membrane.RTCP.TransportFeedbackPacket.TWCC
   alias Membrane.RTP
   alias Membrane.RTP.Header
@@ -62,7 +63,7 @@ defmodule Membrane.RTP.TWCCSender do
   def handle_tick(
         :bandwidth_report_timer,
         _ctx,
-        %{cc: %CongestionControl{last_r_hat: nil}} = state
+        %{cc: %CongestionControl{r_hat: %ReceiverRate{value: nil}}} = state
       ) do
     # wait until first r_hat is calculated
     {:ok, state}
--- a/lib/membrane/rtp/twcc_sender.ex
+++ b/lib/membrane/rtp/twcc_sender.ex
@@ -59,6 +59,16 @@ defmodule Membrane.RTP.TWCCSender do
   end
 
   @impl true
+  def handle_tick(
+        :bandwidth_report_timer,
+        _ctx,
+        %{cc: %CongestionControl{last_r_hat: nil}} = state
+      ) do
+    # wait until first r_hat is calculated
+    {:ok, state}
+  end
+
+  @impl true
   def handle_tick(:bandwidth_report_timer, _ctx, %{cc: cc} = state) do
     {{:ok, notify: {:bandwidth_estimation, min(cc.a_hat, cc.as_hat)}}, state}
   end
--- a/lib/membrane/rtp/twcc_sender.ex
+++ b/lib/membrane/rtp/twcc_sender.ex
@@ -7,8 +7,7 @@ defmodule Membrane.RTP.TWCCSender do
 
   require Bitwise
 
-  alias __MODULE__.CongestionControl
-  alias __MODULE__.ReceiverRate
+  alias __MODULE__.{CongestionControl, ReceiverRate}
   alias Membrane.RTCP.TransportFeedbackPacket.TWCC
   alias Membrane.RTP
   alias Membrane.RTP.Header
--- a/lib/membrane/rtp/twcc_sender.ex
+++ b/lib/membrane/rtp/twcc_sender.ex
@@ -65,7 +65,7 @@ defmodule Membrane.RTP.TWCCSender do
         _ctx,
         %{cc: %CongestionControl{r_hat: %ReceiverRate{value: nil}}} = state
       ) do
-    # wait until first r_hat is calculated
+    # wait until the first r_hat is calculated
     {:ok, state}
   end
 
--- a/lib/membrane/rtp/twcc_sender/congestion_control.ex
+++ b/lib/membrane/rtp/twcc_sender/congestion_control.ex
@@ -54,7 +54,7 @@ defmodule Membrane.RTP.TWCCSender.CongestionControl do
     # latest receiver-side incoming bitrates when we were in the decrease state
     decrease_r_hats: [],
     # latest receiver-side incoming bitrate
-    last_r_hat: 0.0,
+    last_r_hat: nil,
     # time window for measuring the received bitrate, between [0.5, 1]s (reffered to as "T" in the RFC)
     target_receive_window: Time.milliseconds(750),
     # accumulator for packets and their timestamps that have been received through target_receive_interval
@@ -76,7 +76,7 @@ defmodule Membrane.RTP.TWCCSender.CongestionControl do
           a_hat: float(),
           as_hat: float(),
           decrease_r_hats: [float()],
-          last_r_hat: float(),
+          last_r_hat: float() | nil,
           packets_received: [{Membrane.Time.t(), non_neg_integer()}]
         }
 
@@ -251,17 +251,33 @@ defmodule Membrane.RTP.TWCCSender.CongestionControl do
 
   defp update_state(cc, _signal), do: cc
 
+  defp update_receiver_bitrate(
+         %__MODULE__{last_r_hat: nil} = cc,
+         reference_time,
+         receive_deltas,
+         packet_sizes
+       ) do
+    %__MODULE__{target_receive_window: receive_window} = cc
+
+    packets_received = resolve_receive_deltas(receive_deltas, reference_time, packet_sizes)
+
+    packets_received = cc.packets_received ++ packets_received
+
+    {first_packet_timestamp, _first_packet_size} = List.first(packets_received)
+    {last_packet_timestamp, _last_packet_size} = List.last(packets_received)
+
+    if last_packet_timestamp - first_packet_timestamp >= receive_window do
+      cc = %__MODULE__{cc | last_r_hat: 0.0, packets_received: packets_received}
+      update_receiver_bitrate(cc, reference_time, receive_deltas, packet_sizes)
+    else
+      %__MODULE__{cc | packets_received: packets_received}
+    end
+  end
+
   defp update_receiver_bitrate(cc, reference_time, receive_deltas, packet_sizes) do
     %__MODULE__{target_receive_window: receive_window} = cc
 
-    packets_received =
-      receive_deltas
-      |> Enum.zip(packet_sizes)
-      |> Enum.filter(fn {delta, _size} -> delta != :not_received end)
-      |> Enum.scan({reference_time, List.first(packet_sizes)}, fn {recv_delta, size},
-                                                                  {prev_timestamp, _prev_size} ->
-        {prev_timestamp + recv_delta, size}
-      end)
+    packets_received = resolve_receive_deltas(receive_deltas, reference_time, packet_sizes)
 
     {last_packet_timestamp, _last_packet_size} = List.last(packets_received)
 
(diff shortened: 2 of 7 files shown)


lib/membrane/rtp/twcc_sender/congestion_control.ex
```diff
@@ -51,16 +51,10 @@ defmodule Membrane.RTP.TWCCSender.CongestionControl do
...
-    # time window for measuring the received bitrate, between [0.5, 1]s (reffered to as "T" in the RFC)
-    target_receive_interval: Time.milliseconds(750),
-    # accumulator for packet sizes in bits that have been received through target_receive_interval
-    packet_received_sizes: 0,
-    # starting timestamp for current packet received interval
-    packet_received_interval_start: nil,
-    # last timestamp for current packet received interval
-    packet_received_interval_end: nil,
+    # latest receiver-side incoming bitrates when we were in the decrease state
+    decrease_r_hats: [],
```
bblaszkow06, line 55: This could also be `:queue`, you always add one element at head, but then use `Enum.take` to drop the last element which is not very efficient

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,66 @@
...
+    %__MODULE__{rr | value: value, packets_received: packets_received}
+  end
+
+  defp resolve_receive_deltas(receive_deltas, reference_time, packet_sizes) do
+    receive_deltas
+    |> Enum.zip(packet_sizes)
+    |> Enum.filter(fn {delta, _size} -> delta != :not_received end)
+    |> Enum.scan({reference_time, List.first(packet_sizes)}, fn {recv_delta, size},
+                                                                {prev_timestamp, _prev_size} ->
+      {prev_timestamp + recv_delta, size}
```
bblaszkow06, line 63: Since you don't need size to be passed in `acc` I'd use `map_reduce`
```suggestion
    |> Enum.map_reduce(reference_time, fn {recv_delta, size}, prev_timestamp ->
      receive_timestamp = prev_timestamp + recv_delta
      {{receive_timestamp, size}, receive_timestamp}
```

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,66 @@
...
+
+    treshold = last_packet_timestamp - rr.window
+
+    packets_received =
+      (rr.packets_received ++ packets_received)
+      |> Enum.drop_while(fn {timestamp, _size} -> timestamp < treshold end)
+
+    packet_received_sizes = Enum.map(packets_received, fn {_timestamp, size} -> size end)
+
+    value = 1 / (Time.as_milliseconds(rr.window) / 1000) * Enum.sum(packet_received_sizes)
```
bblaszkow06, line 52: You can extract sizes and sum them up in one pass
```suggestion
    received_sizes_sum = Enum.reduce(packets_received, 0, fn {_timestamp, size}, acc -> acc + size end)

    value = 1 / (Time.as_milliseconds(rr.window) / 1000) * received_sizes_sum
```

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,66 @@
...
+  # received by a receiver in last `window` time
+
+  alias Membrane.Time
+
+  @type t() :: %__MODULE__{
+          value: float() | nil,
+          # time window for measuring the received bitrate, between [0.5, 1]s (reffered to as "T" in the RFC)
+          window: Time.t(),
+          # accumulator for packets and their timestamps that have been received through target_receive_interval
+          packets_received: [{Time.t(), pos_integer()}]
```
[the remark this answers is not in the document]
mickel8: The `List.last(packets_received)` in second `update` clause will be faster but on the other hand we had to switch from `Enum.drop_while` to `:queue.filter` and the `:queue` API is pretty wierd. Take a look here https://github.com/membraneframework/membrane_rtp_plugin/pull/115/commits/51cee7277bf0ed98725cd5587ef21e9a6615f4db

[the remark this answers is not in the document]
bblaszkow06: How about `Qex` then? It's already a `membrane_core` dependency and implements Elixir protocols, including `Enumerable`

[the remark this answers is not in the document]
mickel8: Looking at how they implement `Enumerable` https://github.com/princemaple/elixir-queue/blob/v0.5.1/lib/impls/enumerable.ex I think we will get similar results to the list implementation as every time we want to iterate over the queue we are converting it to a list at first and converting to a list is O(n) https://github.com/erlang/otp/blob/master/lib/stdlib/src/queue.erl#L89

So to sum up
1. current implementation should be the fastest one
2. Qex and list implementations should have similar performance: (taking last element, drop_while)
O(1) + O(n) +O(x) - Qex
O(n) + O(x) - list

lib/membrane/rtp/twcc_sender/congestion_control.ex
mickel8: Good point but
1. I would have to store queue size on my own as it is needed in line 303
2. the way we calculate variance in `std_dev/1` isn't the most correct one as we might get catastophic cancellation 
3. exp moving avg could be calculated in constant time but to do so we need to resolve the 2nd point at first as variance can probably be calculated in constant time too
4. I found constant time implementation in Pion https://github.com/pion/interceptor/blob/master/pkg/gcc/rate_controller.go#L33-L48 but honestly speaking I don't fully get how they calculate variance

To sum up,  this code needs to be refactored a little bit anyway so maybe we could do this in a separate PR and don't introduce queue at the moment :stuck_out_tongue:

mickel8: @Observer46 helped me a lot and we found similar implementation in Chromium

https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/link_capacity_estimator.cc;drc=27e39fa1dd71076618c358639ed8a327bc3873c4;bpv=0;bpt=1;l=45

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -58,9 +60,12 @@ defmodule Membrane.RTP.TWCCSender.ReceiverRate do
     receive_deltas
     |> Enum.zip(packet_sizes)
     |> Enum.filter(fn {delta, _size} -> delta != :not_received end)
-    |> Enum.scan({reference_time, List.first(packet_sizes)}, fn {recv_delta, size},
-                                                                {prev_timestamp, _prev_size} ->
-      {prev_timestamp + recv_delta, size}
+    |> Enum.reduce({reference_time, :queue.new()}, fn {recv_delta, size},
+                                                      {prev_timestamp, packets_received} ->
+      receive_timestamp = prev_timestamp + recv_delta
+      {receive_timestamp, :queue.in({receive_timestamp, size}, packets_received)}
```
bblaszkow06, line 66: Since you operate on a list either way, maybe it would more readable to use map_reduce and (after `elem`) using `:queue.from_list`?

lib/membrane/rtp/twcc_sender/receiver_rate.ex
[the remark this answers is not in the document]
bblaszkow06: Well, both `:queue` and `Qex` could use a recursive function implementing the `drop_while` without step O(n).

> current implementation should be the fastest one

I wouldn't be so sure as `filter` is also O(n), so it would all depend on whether O(x) is closer to O(1) or O(n) - and by this code, I was assuming the former.
Anyway, I wouldn't be surprised if all of our estimations were wrong, as most of the complexities for `:queue` are amortized and might degrade due to the way the calls are ordered, not to mention that for only 25 elements the big O complexity might not be that relevant.

So from my perspective, unless benchmarks tip the scale in favour of one of the options, I'd choose the most readable one. I felt that `packets_received` was a FIFO, and API-wise `Qex` > `:queue`, so this is what I would choose, but the difference seems marginal now and I'll accept any option. :wink:

lib/membrane/rtp/twcc_sender/receiver_rate.ex
mickel8: I have changed that but the main difference was that I was creating final queue in one pass so in O(n) while now it is created in 2 * O(n) as :queue.from_list/Qex.new is O(n)

test/membrane/rtp/twcc_sender/congestion_control_test.exs
```diff
@@ -245,8 +245,7 @@ defmodule Membrane.RTP.TWCCSender.CongestionControlTest do
         )
 
       assert cc.state == :decrease
-      assert cc.a_hat >= 0.80 * rate
-      assert cc.a_hat <= 0.95 * rate
+      assert_in_delta cc.a_hat / rate, 0.85, 0.95
```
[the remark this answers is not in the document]
mickel8: Consider such an assertion

```elixir
   assert_in_delta 0.75, 0.8, 0.5
 ```
 
 It will pass, so I would have to write `assert_in_delta, x, 0.875, 0.75` which I think is less redable as we operate on strange `0.875`. The requirement to be between 0.8 and 0.95 is from the RFC draft.
 
 ```
    When an over-use is detected the system transitions to the decrease
   state, where the delay-based available bandwidth estimate is
   decreased to a factor times the currently incoming bitrate.

     A_hat(i) = beta * R_hat(i)

   beta is typically chosen to be in the interval [0.8, 0.95], 0.85 is
   the RECOMMENDED value.
   ```

File: lib/membrane/rtp/twcc_sender/receiver_rate.ex [hunk 7]
```diff
@@ -0,0 +1,73 @@
...
+      %__MODULE__{rr | packets_received: packets_received}
+    end
+  end
+
+  def update(%__MODULE__{} = rr, reference_time, receive_deltas, packet_sizes) do
+    packets_received = resolve_receive_deltas(receive_deltas, reference_time, packet_sizes)
+
+    {last_packet_timestamp, _last_packet_size} = Qex.last!(packets_received)
+
+    treshold = last_packet_timestamp - rr.window
```
balins, line 45:
```suggestion
    threshold = last_packet_timestamp - rr.window
```

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,73 @@
...
+    packets_received = resolve_receive_deltas(receive_deltas, reference_time, packet_sizes)
+
+    packets_received = Qex.join(rr.packets_received, packets_received)
+
+    {first_packet_timestamp, _first_packet_size} = Qex.first!(packets_received)
+    {last_packet_timestamp, _last_packet_size} = Qex.last!(packets_received)
+
+    if last_packet_timestamp - first_packet_timestamp >= rr.window do
+      rr = %__MODULE__{rr | value: 0.0}
+      update(rr, reference_time, receive_deltas, packet_sizes)
```
balins, line 34: I know it will happen only at the beginning, but we could easily extract everything after line 45 in `update` to eg. `do_update` function that will get already resolved receive deltas and `threshold` as arguments to avoid calculating it twice here.

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,73 @@
+defmodule Membrane.RTP.TWCCSender.ReceiverRate do
+  @moduledoc false
+  # module responsible for calculating bitrate
+  # received by a receiver in last `window` time
+  # (referred to as R_hat in the GCC draft, sec. 5.5)
+
+  alias Membrane.Time
+
+  @type t() :: %__MODULE__{
+          value: float() | nil,
```
balins, line 10:
```suggestion
          # estimated bitrate in bps
          value: float() | nil,
```

lib/membrane/rtp/twcc_sender.ex
```diff
@@ -8,6 +8,7 @@ defmodule Membrane.RTP.TWCCSender do
   require Bitwise
 
   alias __MODULE__.CongestionControl
+  alias __MODULE__.ReceiverRate
```
balins, line 11:
```suggestion
  alias __MODULE__.{CongestionControl, ReceiverRate}
```

lib/membrane/rtp/twcc_sender/receiver_rate.ex
```diff
@@ -0,0 +1,73 @@
...
+
+    value = 1 / (Time.as_milliseconds(rr.window) / 1000) * received_sizes_sum
+
+    %__MODULE__{rr | value: value, packets_received: packets_received}
+  end
+
+  defp resolve_receive_deltas(receive_deltas, reference_time, packet_sizes) do
+    receive_deltas
+    |> Enum.zip(packet_sizes)
+    |> Enum.filter(fn {delta, _size} -> delta != :not_received end)
```
balins, line 64:
```suggestion
    |> Enum.reject(fn {delta, _size} -> delta == :not_received end)
```
If it comes to filtering with `!=`, I'm a `reject` enjoyer.
But I don't insist

balins (commented):
hello, darkness, my old friend...

test/membrane/rtp/twcc_sender/congestion_control_test.exs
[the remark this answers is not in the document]
bblaszkow06: You're right, the explicit comparison is more readable since the values come from RFC. I guess you can add a comment referencing the part you quoted to make these values less magical

bblaszkow06 (approved):

balins (approved):

PR #116, opened 2022-10-23
[NOJIRA] Fix `Serializer` and `OutboundPacketTracker` caps

mickel8:


lib/membrane/rtp/outbound_packet_tracker.ex
```diff
@@ -11,18 +11,23 @@ defmodule Membrane.RTP.OutboundPacketTracker do
...
-  alias Membrane.{Buffer, Payload, RTCPEvent, RTP, Time}
+  alias Membrane.{Buffer, Payload, RemoteStream, RTCPEvent, RTP, Time}
 
-  def_input_pad :input, caps: :any, demand_mode: :auto
+  def_input_pad :input, caps: RTP, demand_mode: :auto
 
-  def_output_pad :output, caps: :any, demand_mode: :auto
+  def_output_pad :output,
+    caps: {RemoteStream, type: :packetized, content_format: RTP},
+    demand_mode: :auto
```
bblaszkow06, line 20: It doesn't seem really intuitive to have `Tracker` that changes caps, especially since it doesn't modify the packets.
I think keeping it as RTP would allow SRT(C)P encoders to use RTP caps as well.

lib/membrane/rtp/serializer.ex
```diff
@@ -10,16 +10,14 @@ defmodule Membrane.RTP.Serializer do
...
 
   @max_seq_num Bitwise.bsl(1, 16) - 1
   @max_timestamp Bitwise.bsl(1, 32) - 1
 
   def_input_pad :input, caps: RTP, demand_mode: :auto
 
-  def_output_pad :output,
-    caps: {RemoteStream, type: :packetized, content_format: RTP},
-    demand_mode: :auto
+  def_output_pad :output, caps: RTP, demand_mode: :auto
```
bblaszkow06, line 20: Such change kind of breaks the concept of caps - in theory, the `RemoteStream` has all the headers in the payload and isn't parsed while RTP contains just RTP payload as `Membrane.Buffer`'s payload. This is also symmetrical to the `RTP.Parser` that does the opposite conversion and changes caps from `RemoteStream` to `RTP`.
So the question is - what is the purpose of this change? I may have missed some conversation about it, but if the `RTP` caps are needed here for some reason, we may need more complex changes to the caps inside the RTP plugin. Maybe we should eliminate RemoteStream caps (at least for output) and have some distinction of payloads as a field inside `RTP` caps? Although I guess this solution was rejected at some point when adding RemoteCaps.

bblaszkow06:
OK, I've taken a look at the code outside of the PR and it all makes sense, but since it might take a while until this gets refactored, I'd try to make a step in the right direction by changing the module names. It doesn't seem to break much as the elements are mostly used via bins from the same package. 

I'd suggest:
* `Serializer` -> `HeaderGenerator`, `(Header)MetadataGenerator` or similar.
* `OutboundPacketTracker` is more tricky, but my ideas are: `(Outbound)TrackingSerializer` or just `OutboundSerializer`

With that changes in place, I'll sleep better awaiting a rewrite of the outbound RTP pipeline :wink:

mickel8:
I've introduced requested changes but honestly speaking I would keep `OutboundPacketTracker`. It looks like it is able to generate some parts of the header too :exploding_head: and the previous name was symmetric to `InboundPacketTracker`. I made renamings in separate commits so we can revert the last one if you want to.

PR #118, opened 2022-10-28
Support padding packets

daniel-jodlos:
This PR adds support for generation of padding packets to RTP Plugin. This change is required to be able to publish a version of Membrane RTC Engine with automatic variant switching, as it needs to be able to probe the connection using padding packets.

Out of necessity, this PR also fixes `Membrane.RTP.Utils.align/2` function - it was incorrectly calculating the size of required padding as a `size % align_to` instead of `align_to - (size % align_to)`

### DIFF
--- a/lib/membrane/rtp/outbound_tracking_serializer.ex
+++ b/lib/membrane/rtp/outbound_tracking_serializer.ex
@@ -13,6 +13,8 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
   alias Membrane.RTP.Session.SenderReport
   alias Membrane.{Buffer, Payload, RemoteStream, RTCPEvent, RTP, Time}
 
+  @padding_packet_size 256
+
   def_input_pad :input, caps: RTP, demand_mode: :auto
 
   def_output_pad :output,
@@ -163,6 +165,12 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
         %{extension | identifier: Map.fetch!(state.extension_mapping, extension.identifier)}
       end)
 
+    is_padding_packet? = Map.get(rtp_metadata, :is_padding?, false)
+
+    if is_padding_packet? and buffer.payload != <<>> do
+      raise "Incorrect padding packet. Padding packets must have empty payload"
+    end
+
     header =
       struct(RTP.Header, %{
         rtp_metadata
@@ -173,7 +181,7 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
 
     payload =
       RTP.Packet.serialize(%RTP.Packet{header: header, payload: buffer.payload},
-        align_to: state.alignment
+        align_to: if(is_padding_packet?, do: @padding_packet_size, else: state.alignment)
       )
 
     buffer = %Buffer{buffer | payload: payload, metadata: metadata}
--- a/lib/membrane/rtp/utils.ex
+++ b/lib/membrane/rtp/utils.ex
@@ -24,7 +24,8 @@ defmodule Membrane.RTP.Utils do
       0 ->
         {payload, 0}
 
-      padding_size ->
+      remainder ->
+        padding_size = align_to - remainder
         zeros_no = padding_size - 1
         {<<payload::binary, 0::size(zeros_no)-unit(8), padding_size>>, padding_size}
     end
--- a/test/membrane/rtp/outbound_tracking_serializer_test.exs
+++ b/test/membrane/rtp/outbound_tracking_serializer_test.exs
@@ -1,0 +1,31 @@
+defmodule Membrane.RTP.OutboundTrackingSerializerTest do
+  use ExUnit.Case, async: true
+
+  alias Membrane.Buffer
+  alias Membrane.RTP.OutboundTrackingSerializer
+
+  test "generates padding packets" do
+    serializer_state = %OutboundTrackingSerializer.State{}
+
+    buffer = %Buffer{
+      metadata: %{
+        rtp: %{
+          is_padding?: true,
+          ssrc: 1,
+          payload_type: 0,
+          marker: true,
+          extensions: [],
+          csrcs: [],
+          timestamp: 0,
+          sequence_number: 0
+        }
+      },
+      payload: <<>>
+    }
+
+    assert {{:ok, buffer: {:output, %Buffer{payload: payload}}}, _state} =
+             OutboundTrackingSerializer.handle_process(:input, buffer, nil, serializer_state)
+
+    assert byte_size(payload) == 256
+  end
+end


test/membrane/rtp/outbound_tracking_serializer_test.exs
```diff
@@ -0,0 +1,33 @@
...
+          sequence_number: 0
+        }
+      },
+      payload: <<>>
+    }
+
+    assert {{:ok, buffer: {:output, %Buffer{payload: payload}}}, _state} =
+             OutboundTrackingSerializer.handle_process(:input, buffer, nil, serializer_state)
+
+    assert byte_size(payload) == 256
```
mickel8, line 29: We should check last byte against correct value - it should denote padding size

lib/membrane/rtp/outbound_tracking_serializer.ex
```diff
@@ -13,6 +13,8 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
   alias Membrane.RTP.Session.SenderReport
   alias Membrane.{Buffer, Payload, RemoteStream, RTCPEvent, RTP, Time}
 
+  @padding_packet_size 256
```
mickel8, line 16: I think this should be 255 as the max padding size is denoted by last octet of RTP packet payload

test/membrane/rtp/outbound_tracking_serializer_test.exs
```diff
@@ -0,0 +1,33 @@
+defmodule Membrane.RTP.OutboundTrackingSerializerTest do
```
mickel8, line 1: I think that we should test `RTP.Packet.serialize` instead

mickel8 (changes requested):
Could you please include in this PR fixes for GCC? I remember that padding packets are not counted in GCC algorithm correctly

test/membrane/rtp/outbound_tracking_serializer_test.exs
daniel-jodlos: In this case, yes. However, now that I think about it, we can calculate the size of the header in this particular example, so we can just verify the entire padding binary

lib/membrane/rtp/outbound_tracking_serializer.ex
daniel-jodlos: It's not a big probem, the header will certainly be bigger than 1 byte, but sure, this would make more sense. We could also consider using 255 + min header size

test/membrane/rtp/outbound_tracking_serializer_test.exs
daniel-jodlos: I thought about it as well, but the thing is, `serialize/2` has no clue that it's creating a padding packet, it simply does the alignment to `@padding_packet_size`. For that reason, I wanted to hit `handle_process` to ensure that it detects a padding packet and calls the serialize function accordingly, especially that not working probing would be really bad to all our WebRTC products

lib/membrane/rtp/outbound_tracking_serializer.ex
mickel8: I missunderstood the code at the first time. I thought that `@padding_packet_size` is the length of RTP packet payload. In such a case, 256 would result in 0 in last byte denoting padding size as we would have an overflow.

Now I see that we never get padding packet that has 255 zeros. Instead we always get packet that is 255 bytes long in total, including header.

I think we should fix this. If we want to create padding packet it should have 255 bytes + 1 byte for padding size + header

lib/membrane/rtp/twcc_sender.ex
```diff
@@ -123,10 +123,17 @@ defmodule Membrane.RTP.TWCCSender do
     buffer =
       Header.Extension.put(buffer, %Header.Extension{identifier: :twcc, data: <<seq_num::16>>})
 
+    payload_size =
+      if Map.get(buffer.metadata.rtp, :is_padding?, false) do
+        8 * 255
+      else
+        bit_size(buffer.payload)
+      end
```
mickel8, line 131: What is acctually in `buffer.payload` here? If the whole serialized RTP packet then `bit_size(buffer.payload)` should work for padding packets too

lib/membrane/rtp/outbound_tracking_serializer.ex
```diff
@@ -173,7 +181,7 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
 
     payload =
       RTP.Packet.serialize(%RTP.Packet{header: header, payload: buffer.payload},
-        align_to: state.alignment
+        align_to: if(is_padding_packet?, do: @padding_packet_size, else: state.alignment)
```
mickel8, line 184: I would leave this as it was and handle padding packets inside Packet.serialize. This way we will be able to create padding packets that are 255 + 1 + header bytes long in total and we can focus on testing Packet.serialize instead of OutboundSerializingTracker

mickel8 (changes requested):

bblaszkow06 (commented):
Nothing to add on top of Michał's comments, dismissing a review request until those are fixed.

lib/membrane/rtp/outbound_tracking_serializer.ex
daniel-jodlos: > Padding packet should have 255 + 1 byte for padding size

Nitpick: 255 bytes of padding, period. Padding size is part of padding

lib/membrane/rtp/twcc_sender.ex
daniel-jodlos: I'm pretty sure this is before the serializer, so payload for padding packets is empty. Actually, now that you pointed it out, shouldn't we count the header as well!? I guess for most packet it will be inconsequential, but leaving it as it is seems a little... wrong.

mickel8: I am not sure if we can do anything with this. I would leave `FIXME` tag with the explanation. The whole code will work more consistently anyway

test/membrane/rtp/packet_test.exs
```diff
@@ -90,4 +90,41 @@ defmodule Membrane.RTP.PacketTest do
...
+      payload: <<>>,
+      header: Fixtures.sample_header()
+    }
+
+    serialized = Packet.serialize(packet, is_padding_packet?: true)
+
+    assert {:ok, %{has_padding?: true, packet: %Packet{payload: <<>>}}} =
+             Packet.parse(serialized, false)
+
+    # expected size of the padding packet - the size of the header
```
mickel8, line 105: this comment is not accurate

mickel8 (commented):
I took a look at how the `align_to` option is used and I think we could get rid of it.

`align_to` is used in 2 places, in PayloaderBin, which passes this option to the Serializer and in the OutboundTrackingSerializer. OTS is created only in one place, in SessionBin, and there is no `align_to` option passed to it.

Having said that, it looks like `align_to` could be safely deleted from OTS. When it comes to the PayloaderBin, I think that such alignment should be the responsibility of RTP payloader. 

To sum up, I think that we should revisit `align_to` option and decide whether to get rid of it or not. 

If we decide to keep `align_to` I would ignore it when `is_padding` otpion is set.

When it comes to the `is_padding` option. I believe that in rtp metadata  map, there are only fields from RTP.Header struct. Therefore, allowing for adding a new field to this metadata that doesn't correspond to any field in RTP.Header struct sounds like scripting. We have to either document this option and its behavior or go for another approach.

The another approach is to add `has_padding` flag to the RTP.Header struct and rtp metadata map and output RTP packets with padding as they are. This way we can output packets with padding without modification and we can also provide packets on our own.
RTP plugin would stop being responsible for handling paddings and padding packets.

daniel-jodlos:
> The another approach is to add has_padding flag to the RTP.Header struct and rtp metadata map and output RTP packets with padding as they are. Outputting RTP packets with padding, we could easily create padding packets with has_padding flag set to true and feed them into the RTP plugin. This way RTP plugin would stop being responsible for handling paddings and padding packets.

I was thinking about that, but I bumped straight into the `align_to` option. Because of the way this option works:
- make sure that the byte size of the *entire* packet (including the header) modulo `align_to` == 0
- modify the header to add the padding flag
- add the necessary amount of padding
I don't think it can be applied anywhere else without doing something doggy.

But... maybe we could just kick `align_to` to another function 🤔 We would end up with the following
```elixir
Packet.serialize(Packet.t())
Packet.align_to(<<_version::2, 0::1, _rest::bits>> = serialized_packet, alignment)
```
- OutboundTrackingSerializer looses `align_to` option
- `Packet.align_to` is called manually wherever this is necessary
- `has_padding` would be indicated in `Packet.Header.t()`. I'm still on the fence if it should be a `boolean()`, or maybe `non_neg_integer()` - size of the padding. I'm not very fond of the idea of creating even the padding structure outside of RTP Plugin as it would split the responsibility across multiple places, so I think I would gravitate towards the latter.

mickel8:
Assuming we have to keep `align_to`, separating it from `serialize` sounds good to me. I would just call it before `serialize`.
When it comes to the `has_padding`, I would prefer option where user is responsible for creating padding packets but I don't insist. 

Those ideas need @bblaszkow06 approval before implementing as they modify API significantly.

daniel-jodlos:
> Assuming we have to keep align_to, separating it from serialize sounds good to me. I would just call it before serialize.

Given that `align_to` has to modify the header, I wouldn't. Unless you add `is_padding` to `RTP.Packet.Header` struct, in which case, I agree.

mickel8:
I entirely deleted `align_to` option as:
* it wasn't even exported in SessionBin options. The only way to use it was when someone was spawning PayloaderBin on its own - this means no one was using this option
* we don't even know how this option should work. IMO it should align only payload as only payload is encrypted
* it makes code more complicated

IMO we should bring it back once we need it

mickel8:
I haven't written tests yet but I think that the approach where we send an event to generate padding packet is very inconvenient.

* there is code duplication as we have to handle the event and casual buffers almost in the same way
* it's hard to determine where the correctness of the new event should be checked - ignore event if size is > 255

Therefore, I don't see any benefits from using an event for generating padding packet. Creating payload for padding packet can be done in one line `<<0::size(zeros_no)-unit(8), zeros_no>>` and if there is really someone who wants to generate padding packets, it won't care wheater it has to send an event or a buffer.

In case of SFU, sending an event is worse as in general we operate on RTP packets (munging) but when we want to send a padding packet we have to create event, use event action, have different specs etc.

daniel-jodlos:
I also don't like the event option. The level of duplication is severe and the fields of the event feel like a partial typespec of RTP Buffer's metadata. If we kicked the `align_to` option, I would consider adding `has_padding?` to RTP metadata and just sending the Buffer with appropriate metadata and only padding in the payload from the SFU.

We could also add the `create_padding` function in `Membrane.RTP` to assist in the creation of these buffers 🤔

bblaszkow06 (commented):
Dismissing a review for now

lib/membrane/rtp/packet.ex
```diff
@@ -27,28 +27,18 @@ defmodule Membrane.RTP.Packet do
 
   def identify(_packet), do: :rtp
 
-  @spec serialize(t, align_to: pos_integer()) :: binary
-  def serialize(%__MODULE__{} = packet, [align_to: align_to] \\ [align_to: 1]) do
+  @spec serialize(t, has_padding: boolean()) :: binary
```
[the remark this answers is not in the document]
mickel8: The spec in form of `has_padding: boolean()` can be also written as `[{:has_padding, boolean}]` which means that this is a list of zero or more elements so an empty list passes here.

I would have to write `[type, ...]` to make it not passing.

lib/membrane/rtp/utils.ex
```diff
@@ -17,6 +17,21 @@ defmodule Membrane.RTP.Utils do
...
   end
 
+  @doc """
+  Generates padding binary.
+
+  `padding_size` is the final padding size including
+  the last byte denoting padding size. `padding_size`
+  cannot be greater than 255.
+  """
+  @spec generate_padding(non_neg_integer()) :: binary()
```
daniel-jodlos, line 27: Let's include padding size constraints in the spec
```suggestion
  @spec generate_padding(0..255) :: binary()
```

lib/membrane/rtp/session_bin.ex
```diff
@@ -36,6 +36,25 @@ defmodule Membrane.RTP.SessionBin do
...
+
+  To send a packet with a padding, one should include `padding_size` field in `Membrane.Buffer`'s metadata.
+
+  E.g.
+
+  ```elixir
+  metadata = %{rtp: %{padding_size: 20}}
+  ```
+
+  will result in adding 20 bytes of padding (in total i.e. including the last byte denoting padding size)
```
daniel-jodlos, line 50: Isn't it a little too specific? This last byte *is* part of the padding.

```suggestion
  will result in adding 20 bytes of padding
```

daniel-jodlos (commented):
I cannot approve the PR I created. That said, @mickel8 needs to approve it because it blocks merging.

test/support/rtcp_fixtures.ex
```diff
@@ -1,22 +1,11 @@
 defmodule Membrane.RTCP.Fixtures do
   @moduledoc false
 
-  @external_resource "test/fixtures/rtcp/single_packet.hex"
-  @sample_rtcp_packet File.read!(@external_resource)
-
-  @external_resource "test/fixtures/rtcp/packets.hex"
```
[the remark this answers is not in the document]
bblaszkow06: It shouldn't be deleted as it instructs the compiler to depend on that file, i.e. module `Membrane.RTCP.Fixtures` should be recompiled in case any of these files change.
See https://hexdocs.pm/elixir/Module.html#module-external_resource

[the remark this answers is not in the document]
mickel8: Good to know, I restored the old version

lib/membrane/rtp/utils.ex
bblaszkow06: I'd define a type `padding_size` somewhere and reference it here as it is also used in the spec of `Packet.serialize` and the result type of `Packet.parse`

bblaszkow06 (approved):

PR #122, opened 2022-11-23
[RTC-29] Update the plugin to use new version of `ex_libsrtp`

daniel-jodlos:
This Pull Request bumps the version of `ex_libsrtp` to `v0.6.0` and changes existing code to handle new errors.

Some errors still result in an exception being thrown. In particular, all errors described as "an error in cryptographic mechanisms" result in an Exception, except for `:replay_fail` and `:replay_old`, which we chose to ignore as retransmissions can cause them.


lib/membrane/srtcp/decryptor.ex
```diff
@@ -46,8 +46,22 @@ if Code.ensure_loaded?(ExLibSRTP) do
 
     @impl true
     def handle_process(:input, buffer, _ctx, state) do
-      {:ok, payload} = ExLibSRTP.unprotect_rtcp(state.srtp, buffer.payload)
-      {{:ok, buffer: {:output, %Buffer{buffer | payload: payload}}}, state}
+      case ExLibSRTP.unprotect_rtcp(state.srtp, buffer.payload) do
+        {:ok, payload} ->
+          {{:ok, buffer: {:output, %Buffer{buffer | payload: payload}}}, state}
+
+        {:error, reason} when reason in [:replay_fail, :replay_old, :auth_fail] ->
```
bblaszkow06, line 53: I wonder about keeping the `:auth_fail` here - If we receive such a packet it means someone has found an opened port and sent something there. Not terminating such a connection makes us vulnerable to DoS attacks. Are there any other reasons not to raise in such a case?

mickel8 (approved):

lib/membrane/srtcp/decryptor.ex
daniel-jodlos: Not any good ones. Let's just raise here.

bblaszkow06 (approved):

PR #123, opened 2022-11-30
[RTC-43] Add SSRC to vad notifications

daniel-jodlos:
As part of RTC-39 I'm reenabling VAD in Membrane RTC Engine and one of the problems I encountered is that previously we had a silent assumption that each peer will only have one inbound audio track, which isn't necessarily true.
Because of this, I have added SSRC to VAD notifications to be able to differentiate.

### DIFF
--- a/lib/membrane/rtp/session_bin.ex
+++ b/lib/membrane/rtp/session_bin.ex
@@ -635,7 +635,8 @@ defmodule Membrane.RTP.SessionBin do
   end
 
   @impl true
-  def handle_notification({:vad, _val} = msg, _from, _ctx, state) do
+  def handle_notification({:vad, value}, {:vad, ssrc}, _ctx, state) do
+    msg = {:vad, {ssrc, value}}
     {{:ok, notify: msg}, state}
   end
 


lib/membrane/rtp/session_bin.ex
```diff
@@ -635,7 +635,8 @@ defmodule Membrane.RTP.SessionBin do
   end
 
   @impl true
-  def handle_notification({:vad, _val} = msg, _from, _ctx, state) do
+  def handle_notification({:vad, value}, {:vad, ssrc}, _ctx, state) do
+    msg = {:vad, {ssrc, value}}
```
bblaszkow06, line 639: This generates a new notification that is not documented anywhere. I think you should rather modify the message generated from `RTP.VAD`, you can get the ssrc from the buffer metadata

bblaszkow06 (changes requested):

lib/membrane/rtp/session_bin.ex
daniel-jodlos: Actually, it turns out that RTC Engine in fact doesn't need to have a knowledge of Active Speaker in the context of the peer, the context of a single track is sufficient. Therefore, we might send an event instead of this notification.

daniel-jodlos:
It was replaced by #126

PR #124, opened 2022-12-02
Fix stripping padding out when packet is encrypted

mickel8:
Implementing and merging the proper solution might take a few days so to unblock other teams I implemented small workaround. At the same time I started working on converting encryptor and decryptor into Elixir modules.


bblaszkow06 (approved):
LGTM and seems to fix issues for me!

PR #132, opened 2023-01-23
Add gap detection and NACK sending

bblaszkow06:


test/membrane/rtp/sequence_number_tracker_test.exs
```diff
@@ -0,0 +1,68 @@
...
+
+    test "late sequence number" do
+      tracker = Tracker.new()
+      assert {1, 10, tracker} = Tracker.track(tracker, 10)
+      assert {-1, 9, tracker} = Tracker.track(tracker, 9)
+      assert {2, 12, tracker} = Tracker.track(tracker, 12)
+      assert {1, 13, _tracker} = Tracker.track(tracker, 13)
+    end
+
+    test "gap and late at rollover" do
```
mickel8, line 52: This is very specific case as the last seq_num is a duplicate. Would be nice to add sth like 65_533, 65_534, 1, 65_535  or 65_532, 65_533, 1, 65_534

test/membrane/rtp/sequence_number_tracker_test.exs
```diff
@@ -0,0 +1,68 @@
...
+
+  describe "track/2" do
+    test "continuous sequence" do
+      tracker = Tracker.new()
+      assert {1, 10, tracker} = Tracker.track(tracker, 10)
+      assert {1, 11, tracker} = Tracker.track(tracker, 11)
+      assert {1, 12, _tracker} = Tracker.track(tracker, 12)
+    end
+
+    test "numbers starting at 0" do
```
mickel8, line 14: I would also add 0, 65_535, 1 sequence

mickel8 (changes requested):

lib/membrane/rtp/inbound_packet_tracker.ex
```diff
@@ -74,35 +74,29 @@ defmodule Membrane.RTP.InboundPacketTracker do
...
 
-        {[buffer: {:output, repair_sequence_number(buffer, state)}], state}
+    lost_ids = Enum.to_list((packet_index - (delta - 1))..(packet_index - 1)//1)
 
-      # the packets is either too old or too new
-      delta <= @max_seq_num - @max_unordered ->
-        {[], update_received(state)}
+    if abs(delta) > @max_unordered do
+      # Ignore packet
+      {[], state}
```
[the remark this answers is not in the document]
bblaszkow06: I'm not 100% sure, I've just combined the existing conditions. The previous version dropped packets with diff >= 3000 and < (65535 - 3000). I believe this is to rule out a scenario where we receive a terribly late packet and consider it a newer one, breaking the sequence number tracking and causing an enormous amount of NACKs to be sent

The scenario example:
```elixir
test "huge gap" do
  tracker = Tracker.new()
  assert {1, 1, tracker} = Tracker.track(tracker, 1)
  assert {32_767, 32_768, tracker} = Tracker.track(tracker, 32_768)
  assert {-32_766, 2, tracker} = Tracker.track(tracker, 2)
  assert {-32_765, 3, tracker} = Tracker.track(tracker, 3)
end
```

However, as shown above, this will break the `SequenceNumberTracker` anyway.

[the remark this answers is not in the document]
bblaszkow06: Now state update is only done when we're not dropping packets and increased the limit 3x.
I've also added some comments

test/membrane/rtp/sequence_number_tracker_test.exs
bblaszkow06: Done as separate case

mickel8 (approved):

PR #136, opened 2023-02-01
Fix TWCC not being reset when adding the first pad

daniel-jodlos:


lib/membrane/rtp/twcc_sender.ex
```diff
@@ -32,9 +32,25 @@ defmodule Membrane.RTP.TWCCSender do
...
   @impl true
-  def handle_pad_added(pad, _ctx, state) do
+  def handle_pad_added(pad, ctx, state) do
     {queued_actions, other_actions} = Map.pop(state.buffered_actions, pad, [])
 
+    has_other_input_pads? =
+      ctx.pads
+      |> Map.values()
+      |> Enum.count(&(&1.ref != pad && &1.direction == :input))
+      |> then(&(&1 != 0))
```
bblaszkow06, line 42:
```suggestion
      |> Map.values()
      |> Enum.any?(&(&1.ref != pad and &1.direction == :input))
```

or using only one iteration

```suggestion
      |> Enum.any?(fn
        {ref, %{direction: :input}} when ref != pad -> true
        _else -> false
      end)
```

bblaszkow06 (approved):

lib/membrane/rtp/twcc_sender.ex
```diff
@@ -32,9 +32,25 @@ defmodule Membrane.RTP.TWCCSender do
   end
 
   @impl true
-  def handle_pad_added(pad, _ctx, state) do
+  def handle_pad_added(pad, ctx, state) do
```
bblaszkow06, line 35: Oh, w8, one more thing - we should only invoke this when new input pad is added, otherwise we will reset `cc` twice

bblaszkow06 (changes requested):

mickel8 (commented):
To me it looks suspicious. 

I would expect that if we don't send any packets we also won't get any TWCC reports. If we don't get any TWCC reports, the estimation should stay unchanged. 

That's how we update cc

```elixir
    cc =
      CongestionControl.update(
        state.cc,
        reference_time,
        receive_deltas,
        send_deltas,
        packet_sizes,
        rtt
      )
```

where reference time is taken from the last received TWCC report. So we have differences between sent packets, received packets and current rtt. Doesn't look like the time when we created CC struct should have any impact on the estimation :thinking: 

Maybe there is something inside CC struct that makes it time sensitive

daniel-jodlos:
> If we don't get any TWCC reports, the estimation should stay unchanged.

And it does. Until we receive a packet, then it drops to basically nothing.

> Doesn't look like the time when we created CC struct should have any impact on the estimation thinking

It does. `r_hat` calculation uses time window to estimate the receiver rate and it doesn't have its own clock - it relies on RTP and RTCP packets to update.  It makes perfect sense to me, that when you have a 750ms window, and you don't send any packets for a while, and then you send one, rr it estimated to be 4/3 of the size of your packet and the estimation is adjusted accordingly.

bblaszkow06 (approved):

PR #142, opened 2023-02-08
[RTC-93] Add some RTP metrics

daniel-jodlos:
This PR adds a few of the metrics described in RTC-93:
1. Inbound frames (estimated based on the marker)
2. Outbound frames (estimated based on the marker)
2. Inbound RTP NACK counter
3. Inbound RTCP Packets
4. Inbound RTCP Bytes Received
5. Outbound RTCP Packets
6. Outbound RTCP Bytes Sent


lib/membrane/rtp/metrics.ex
```diff
@@ -3,34 +3,66 @@ defmodule Membrane.RTP.Metrics do
...
         measurement: :encoding
       ),
-      Telemetry.Metrics.last_value(
+      Metrics.last_value(
         "inbound-rtp.ssrc",
         event_name: [Membrane.RTP, :inbound_track, :new],
         measurement: :ssrc
+      ),
+      Metrics.counter(
+        "inbound-rtp.nack",
```
bblaszkow06, line 38: Just to be sure, I'd include `_sent` in the name to leave no questions
```suggestion
        "inbound-rtp.nack_sent",
```

lib/membrane/rtcp/parser.ex
```diff
@@ -58,6 +80,14 @@ defmodule Membrane.RTCP.Parser do
   @impl true
   def handle_event(:output, %RTCPEvent{} = event, _ctx, state) do
     buffer = %Buffer{payload: RTCP.Packet.serialize(event.rtcp)}
+
+    Membrane.TelemetryMetrics.execute(
+      @rtcp_sent_telemetry_event,
```
bblaszkow06, line 85: This might be a bit problematic, as it includes FIRs, PLIs, NACKs, REMB (when/if implemented) etc in addition to receiver reports.
1. Do you thing such aggregated statistic is helpful?
2. I guess the number of SRs can be estimated by substraction `RTCP - FIR - NACK`, but maybe we should have a separate metric for that

lib/membrane/rtp/outbound_tracking_serializer.ex
```diff
@@ -65,6 +72,8 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
 
   @impl true
   def handle_init(_ctx, options) do
+    Membrane.TelemetryMetrics.register(@frame_sent_telemetry_event, options.telemetry_label)
```
bblaszkow06, line 75: You are ignoring RTCP packets sent by this element, a metric for that is missing

lib/membrane/rtp/metrics.ex
```diff
@@ -3,34 +3,66 @@ defmodule Membrane.RTP.Metrics do
...
+      ),
+      Metrics.counter(
+        "outbound-rtcp.packets",
+        event_name: [Membrane.RTP, :rtcp, :sent]
+      ),
+      Metrics.sum(
+        "outbound-rtcp.bytes_sent",
+        event_name: [Membrane.RTP, :rtcp, :sent],
+        measurement: :bytes
+      ),
```
bblaszkow06, line 58: The metrics you've added are on a receiver (inbound) end, so I would rename them, as they might be confused with RTCP sent as sender feedback (mostly in OutboundTrackingSerializer for now). I think the pattern should rather be `inbound-rtp.rtcp_sent` or `inbound.rtcp_sent`. This resembles stats in `wertc-internals` where such info is aggregated under `inbount-rtp` and `remote-inbound-rtp` (we may also follow that convention).

bblaszkow06 (changes requested):

lib/membrane/rtcp/parser.ex
```diff
@@ -37,6 +52,13 @@ defmodule Membrane.RTCP.Parser do
 
   @impl true
   def handle_process(:input, %Buffer{payload: payload, metadata: metadata}, _ctx, state) do
+    Membrane.TelemetryMetrics.execute(
+      @rtcp_received_telemetry_event,
+      %{bytes: byte_size(payload)},
```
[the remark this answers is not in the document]
daniel-jodlos: :thinking: 
I can't say I have any idea why that would be useful, but Michał requested it.
cc @mickel8 ?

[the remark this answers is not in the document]
mickel8: The idea is to measure % of RTCP traffic

lib/membrane/rtcp/parser.ex
daniel-jodlos: 1. I think this comment answers it pretty well: https://github.com/membraneframework/membrane_rtp_plugin/pull/142#discussion_r1108397831

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -295,6 +309,19 @@ defmodule Membrane.RTP.SSRCRouter do
     )
   end
 
+  defp emit_inbound_frame_event(actions, ctx) do
+    for {:buffer, {pad, %Buffer{} = buffer}} <- actions,
+        buffer.metadata.rtp.marker,
+        Map.has_key?(ctx.pads, pad) do
```
mickel8, line 315: Instead of checking againts pad in ctx.pads, I would emit event every time we receive some buffer. We could also get rid of `emit_inbound_frame_event` in `handle_pad_added(:output)`.

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -99,6 +100,9 @@ defmodule Membrane.RTP.SSRCRouter do
     register_new_inbound_track_event(pad, ctx)
     emit_new_inbound_track_event(ssrc, pad, ctx)
```
mickel8, line 102: In 98 we have `emit_packet_arrival_events` and it seems to be redundant as we always call `emit_packet_arrival_events` in handle_process.  Could we fix this?

mickel8 (changes requested):
Could you please upload some example report with new metrics?

lib/membrane/rtp/metrics.ex
```diff
@@ -3,34 +3,70 @@ defmodule Membrane.RTP.Metrics do
...
-      Telemetry.Metrics.sum(
+      Metrics.sum(
         "inbound-rtp.bytes_received",
         event_name: [Membrane.RTP, :packet, :arrival],
         measurement: :bytes
       ),
-      Telemetry.Metrics.last_value(
+      Metrics.last_value(
         "inbound-rtp.encoding",
         event_name: [Membrane.RTP, :inbound_track, :new],
```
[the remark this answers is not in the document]
daniel-jodlos: It seemed a little weird to me as well, but I have no idea what this will break.

lib/membrane/rtp/outbound_tracking_serializer.ex
```diff
@@ -186,6 +210,15 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
         |> Enum.map(&Membrane.RTCP.Packet.serialize(&1))
         |> Enum.map(&{:buffer, {rtcp_output, %Membrane.Buffer{payload: &1}}})
 
+      for _buffer_action <- actions do
+        Membrane.TelemetryMetrics.execute(
+          @sender_report_sent_telemetry_event,
```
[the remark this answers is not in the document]
daniel-jodlos: I've spend a good hour making sense of that flow and I *think* it won't go to the RTCP Parser, so it shouldn't be counted twice.

lib/membrane/rtp/ssrc_router.ex
daniel-jodlos: This would make it inconsistent with "inbound-rtp.packets". Effectively, you could include a frame in the metrics even though you didn't count packets that make it up.

lib/membrane/rtp/ssrc_router.ex
daniel-jodlos: No, because this is not a fix, this is a regression. Packets are buffered if they arrive before their output pad is connected. This function call handles buffered packets, it's not redundant.

lib/membrane/rtp/metrics.ex
```diff
@@ -3,34 +3,70 @@ defmodule Membrane.RTP.Metrics do
...
+        measurement: :bytes
+      ),
+      Metrics.counter(
+        "inbound-rtp.frames",
+        event_name: [Membrane.RTP, :rtp, :frame_received]
+      ),
+      Metrics.counter(
+        "outbound-rtp.frames",
+        event_name: [Membrane.RTP, :rtp, :frame_sent]
+      ),
```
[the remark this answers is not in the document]
daniel-jodlos: This is a little bit more low-level metric. Maybe we should even rename this metric to simply `markers`. As far as I'm aware, the meaning of the marker can be different depending on the codec, and in Chrome, `frames/s` doesn't seem to count corrupted frames, I think it's coming from the decoder/encoder.

lib/membrane/rtp/ssrc_router.ex
mickel8: You're right, thought that we always emit an event in handle_process but we do this only when buffer is not cached

daniel-jodlos:
After some discussions, I ended up completely reworking all RTCP metrics. They are now grouped per track and they include: NACK sent/received, PLI received (I don't think we're sending any), FIR sent/received, Receiver Reports sent/received and Sender Reports sent/received.

I also had to make a choice of where to count the specific feedback messages and I decided to count them wherever they are being handled. Eg, FIR is counted in OutboundTrackingSerializer which also sends a Keyframe Request upstream as a result of receiving FIR.  The other option was to count them in one central location, which is where the overall counter is incremented, but it occurred to me that if they are counted wherever they are handled, the metric would indicate that we're loosing the packet somewhere, eg, if we have 100 NACK sent by the browser and 0 received.

Example report generated from this version:
```
Reporter.scrape(VideoRoomReporter) #=> %{
  :"inbound-rtp.frames" => 2856,
  :"inbound-rtp.keyframes" => 14,
  {:room_id, "test"} => %{
    {:peer_id, "44896930-ea5c-43d2-87fd-7401aabd810d"} => %{
      :"ice.binding_requests_received" => 19,
      :"ice.binding_responses_sent" => 19,
      :"ice.bytes_received" => 8120624,
      :"ice.bytes_sent" => 5478743,
      :"ice.packets_received" => 9919,
      :"ice.packets_sent" => 6243,
      :"peer.bandwidth" => 2084164.3378765725,
      :"peer.metadata" => nil,
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:54d55d81-f6c5-4a12-8ea4-b44e649b7396:"} => %{
        "inbound-rtp.bytes_received": 160546,
        "inbound-rtp.encoding": :OPUS,
        "inbound-rtp.markers_received": 1,
        "inbound-rtp.packets": 1477,
        "inbound-rtp.ssrc": 1523586566,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 7,
        "rtcp.total_packets_received": 7,
        "rtcp.total_packets_sent": 122,
        "track.metadata": %{"active" => true, "type" => "audio"}
      },
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:acb17d95-5f2a-46ea-87c8-6b0d1aec8111:h"} => %{
        "inbound-rtp.bytes_received": 5710780,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 700,
        "inbound-rtp.packets": 5166,
        "inbound-rtp.ssrc": 4010095982,
        "rtcp.fir_sent": 1,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 120,
        "rtcp.total_packets_received": 120,
        "rtcp.total_packets_sent": 6,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:acb17d95-5f2a-46ea-87c8-6b0d1aec8111:l"} => %{
        "inbound-rtp.bytes_received": 362315,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 691,
        "inbound-rtp.packets": 744,
        "inbound-rtp.ssrc": 1335899795,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 27,
        "rtcp.total_packets_received": 27,
        "rtcp.total_packets_sent": 5,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:acb17d95-5f2a-46ea-87c8-6b0d1aec8111:m"} => %{
        "inbound-rtp.bytes_received": 1852102,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 698,
        "inbound-rtp.packets": 1866,
        "inbound-rtp.ssrc": 1055108907,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 38,
        "rtcp.total_packets_received": 38,
        "rtcp.total_packets_sent": 5,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:72794a35-c51f-46c2-9f3c-66c668bceb22"} => %{
        "inbound-rtp.ssrc": 2593640056,
        "outbound-rtp.markers_sent": 650,
        "outbound-rtp.variant": :high,
        "outbound-rtp.variant-reason": :other,
        "rtcp.sender_reports_sent": 5,
        "rtcp.total_packets_received": 26,
        "rtcp.total_packets_sent": 5
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:ba9d1b41-2f4e-4d6e-bb1c-7ce28126c1ae"} => %{
        "inbound-rtp.ssrc": 251360988,
        "outbound-rtp.variant": :high,
        "outbound-rtp.variant-reason": :other,
        "rtcp.sender_reports_sent": 5,
        "rtcp.total_packets_received": 6,
        "rtcp.total_packets_sent": 5
      }
    },
    {:peer_id, "92bf850f-ff9b-4b43-825f-d160c91656ca"} => %{
      :"ice.binding_requests_received" => 17,
      :"ice.binding_responses_sent" => 17,
      :"ice.bytes_received" => 7630666,
      :"ice.bytes_sent" => 5567842,
      :"ice.packets_received" => 9288,
      :"ice.packets_sent" => 6406,
      :"peer.bandwidth" => 2084966.4894894434,
      :"peer.metadata" => nil,
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:54d55d81-f6c5-4a12-8ea4-b44e649b7396"} => %{
        "inbound-rtp.ssrc": 3034914721,
        "outbound-rtp.variant": :high,
        "outbound-rtp.variant-reason": :other,
        "rtcp.sender_reports_sent": 5,
        "rtcp.total_packets_received": 5,
        "rtcp.total_packets_sent": 5
      },
      {:track_id,
       "44896930-ea5c-43d2-87fd-7401aabd810d:acb17d95-5f2a-46ea-87c8-6b0d1aec8111"} => %{
        "inbound-rtp.ssrc": 1784617135,
        "outbound-rtp.markers_sent": 660,
        "outbound-rtp.variant": :high,
        "outbound-rtp.variant-reason": :other,
        "rtcp.sender_reports_sent": 5,
        "rtcp.total_packets_received": 25,
        "rtcp.total_packets_sent": 5
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:72794a35-c51f-46c2-9f3c-66c668bceb22:h"} => %{
        "inbound-rtp.bytes_received": 5360183,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 653,
        "inbound-rtp.packets": 4759,
        "inbound-rtp.ssrc": 746097665,
        "rtcp.fir_sent": 1,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 116,
        "rtcp.total_packets_received": 116,
        "rtcp.total_packets_sent": 6,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:72794a35-c51f-46c2-9f3c-66c668bceb22:l"} => %{
        "inbound-rtp.bytes_received": 352955,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 651,
        "inbound-rtp.packets": 712,
        "inbound-rtp.ssrc": 3519910266,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 26,
        "rtcp.total_packets_received": 26,
        "rtcp.total_packets_sent": 5,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:72794a35-c51f-46c2-9f3c-66c668bceb22:m"} => %{
        "inbound-rtp.bytes_received": 1749322,
        "inbound-rtp.encoding": :H264,
        "inbound-rtp.markers_received": 651,
        "inbound-rtp.packets": 1778,
        "inbound-rtp.ssrc": 2961107475,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 38,
        "rtcp.total_packets_received": 38,
        "rtcp.total_packets_sent": 5,
        "track.metadata": %{"active" => true, "type" => "camera"}
      },
      {:track_id,
       "92bf850f-ff9b-4b43-825f-d160c91656ca:ba9d1b41-2f4e-4d6e-bb1c-7ce28126c1ae:"} => %{
        "inbound-rtp.bytes_received": 133465,
        "inbound-rtp.encoding": :OPUS,
        "inbound-rtp.markers_received": 1,
        "inbound-rtp.packets": 1379,
        "inbound-rtp.ssrc": 54295950,
        "rtcp.receiver_reports_sent": 5,
        "rtcp.sender_reports_received": 6,
        "rtcp.total_packets_received": 6,
        "rtcp.total_packets_sent": 115,
        "track.metadata": %{"active" => true, "type" => "audio"}
      }
    }
  }
}
```

Also, making note of some bugs:
* outbound tracks have `inbound-rtp.ssrc` metric and it seems a little weird
* there are 2 metrics that ended up on the top-level and I think they originate from RTC Engine

mickel8 (changes requested):
The report starts looking really nice! Also, great work with RTCP metrics. Now we get so much useful info :eyes: 

A couple of notes:
*  Looking at [RTC-93](https://membraneframework.atlassian.net/browse/RTC-93?atlOrigin=eyJpIjoiYWQ2MGZlMWVkZGNjNDI1NjkzZmE1OWFlMzMyOGFjMGQiLCJwIjoiaiJ9) some metrics are still missing:
    *  RTX has been merged so would be nice to add metric about number of retransmitted packets
    * outgoing RTP packets
    * outgoing RTP bytes
    * incoming RTCP bytes
    * outgoing RTCP bytes
    * incoming padding packets
* when some event doesn't occur its metric is not present in the report at all. Would be nice to initilize metrics with default values
* maybe instead of identifying metrics as inbound or outbound we should identify track as inbound or outbound and keep metric name generic when possible. This way we would have 
```
{:track_id, id} => {
  type: :outbound,
  ssrc: 1234
  packets: 1000,
  fir: 100
}
```

We might consider getting rid of RTCP bytes as we have detailed numbers about specific RTCP messages but I belive that such a metric can still be useful

lib/membrane/rtcp/parser.ex
```diff
@@ -6,6 +6,7 @@ defmodule Membrane.RTCP.Parser do
   use Membrane.Filter
 
   require Membrane.Logger
+  require Membrane.TelemetryMetrics
```
mickel8, line 9: Unneeded

mickel8 (commented):
One more concern. Currently total RTCP packets are counted in multiple places e.g. in SSRCRouter and OutboundTrackingSerializer. Both processes register as owners of the same event and I am not sure if that's correct.

If that's correct, then I think we should go further and increment total RTCP packets in every place that some specific RTCP messag is sent or received. 
With current RTP plugin complexity it's very easy either not to count in some metric in total statistics or count it in two or more times

daniel-jodlos:
> * incoming RTCP bytes

This metric may fall victim to the architecture of the RTP Plugin. This value is known in RTCP Parser, but it has no way of knowing to which track this particular RTCP belongs. SSRC Router can know that information, but it receives the RTCP in a parsed form, so it doesn't know the size :shrug: For symmetry, I didn't include outgoing RTCP bytes

daniel-jodlos:
> Both processes register as owners of the same event and I am not sure if that's correct.

It's fine, we do that in a lot of places. We can only have one owner *process* of the `{metric, label}` pair, but I don't think this causes any problems

mickel8:
Ok, then let's do those two things (metric default value and some renamings) in different PRs. Could you please continue this topic?

lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -61,6 +72,13 @@ defmodule Membrane.RTP.OutboundRtxController do
       Membrane.Logger.debug(
         "Retransmitting #{length(buffers_to_retransmit)} buffer(s): #{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}"
       )
+
+      Membrane.TelemetryMetrics.execute(
+        @retransmission_telemetry_event,
+        %{amount: length(buffers_to_retransmit)},
```
mickel8, line 78: Consider extracting buffers length into a variable as we also use it in l:73

mickel8 (commented):
What about other metrics I mentiond in the previous CR?

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -325,18 +333,28 @@ defmodule Membrane.RTP.SSRCRouter do
...
       end
     end
   end
 
-  defp emit_packet_arrival_event(payload, pad, ctx) do
+  defp emit_packet_arrival_event(%Buffer{} = buffer, pad, ctx) do
+    packet_size = byte_size(buffer.payload)
+
+    rtp_payload_size =
+      packet_size - buffer.metadata.rtp.padding_size - buffer.metadata.rtp.total_header_size
```
mickel8, line 345: That's true only when RTP packet is encrypted. Otherwise, buffer.payload only contains RTP payload 
https://github.com/membraneframework/membrane_rtp_plugin/blob/master/lib/membrane/rtp/packet.ex#L120

I would consider checking against padding_size == 254

lib/membrane/rtp/session_bin.ex
```diff
@@ -695,10 +695,16 @@ defmodule Membrane.RTP.SessionBin do
         struct(Membrane.SRTP.Decryptor, %{policies: state.srtp_policies})
       )
 
-    links_generator = fn twcc? ->
+    links_generator = fn twcc?, ctx ->
+      ssrc_router_pad_options = [
+        encoding: :rtx,
```
mickel8, line 700: Is that needed? It looks like SSRC router doesn't make any use of `encoding`...

lib/membrane/rtp/outbound_tracking_serializer.ex
```diff
@@ -41,13 +41,24 @@ defmodule Membrane.RTP.OutboundTrackingSerializer do
                 default: []
               ]
 
+  @packet_sent_telemetry_event [Membrane.RTP, :packet, :sent]
```
mickel8, line 44: Looks like we don't emit this event

mickel8 (changes requested):

lib/membrane/rtp/session_bin.ex
daniel-jodlos: It does. Only to emit `new_inbound_track_event` but it does

mickel8 (approved):

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -286,13 +305,53 @@ defmodule Membrane.RTP.SSRCRouter do
...
       @packet_arrival_event,
-      %{bytes: byte_size(payload)},
+      %{bytes: packet_size},
       %{},
-      ctx.pads[pad].options.telemetry_label
+      label
     )
+
+    if packet_size == 254,
+      do: Membrane.TelemetryMetrics.execute(@padding_packet_arrival_event, %{}, %{}, label)
```
bblaszkow06, line 354: IIRC this is not the correct way to detect padding packets, it will only work for encrypted packets, regular RTP will have the padding stripped

lib/membrane/rtp/ssrc_router.ex
```diff
@@ -286,13 +305,53 @@ defmodule Membrane.RTP.SSRCRouter do
...
+      ctx.pads[destination].options.telemetry_label
+    )
+  end
+
+  defp emit_packet_arrival_events(actions, ctx) do
+    for action <- actions do
+      with {:buffer, {pad, buffer}} <- action do
+        emit_packet_arrival_event(buffer, pad, ctx)
+      end
+    end
```
bblaszkow06, line 338: 1. `for` filters only matching elements
2. I'd return `:ok` since the result of for is not intended to be used outside
```suggestion
    for {:buffer, {pad, buffer}} <- actions do
      emit_packet_arrival_event(buffer, pad, ctx)
    end
    
    :ok
```

bblaszkow06 (changes requested):

lib/membrane/rtp/ssrc_router.ex
daniel-jodlos: Actually, it will work a little too well :facepalm: It will also count non-padding packets that happen to have that exact size

daniel-jodlos: I think we should drop this metric for now and get this Pull Request through without it.

Given that the size is not enough to determine if the packet is a padding packet **AND** the padding also seems to be encrypted, we need something that lives after the decryptor to detect padding packets.
I'm also not convinced it would be a very useful metric on its own in the RTP plugin outside of WebRTC, so I'm considering implementing it in the RTC Engine. Buffers that pass through RTC Engine are decrypted and they have their headers removed, so it should be a simple matter of `byte_size(packet) == 0 || packet == <<0::binary-size(254), 255::8>>`

lib/membrane/rtcp/receiver.ex
```diff
@@ -78,12 +75,7 @@ defmodule Membrane.RTCP.Receiver do
 
   @impl true
   def handle_event(:input, %RTCPEvent{rtcp: %SenderReportPacket{} = rtcp} = event, _ctx, state) do
-    Membrane.TelemetryMetrics.execute(
-      @sender_report_received_telemetry_event,
-      %{},
-      %{},
-      state.telemetry_label
-    )
+    emit_sender_report_sent_telemetry_event(state)
```
bblaszkow06, line 78: 1. This function name is invalid (it sends `@sender_report_reeceived_telemetry_event`)
2. I rather had this in mind:
```suggestion
    emit_telemetry_event(@sender_report_received, state)
```
This way the private function could extract the common part. Now you still have duplication, although grouped. However, I won't insist on that change

bblaszkow06 (changes requested):

lib/membrane/rtp/ssrc_router.ex
bblaszkow06: >I'm also not convinced it would be a very useful metric on its own in the RTP plugin outside of WebRTC, so I'm considering implementing it in the RTC Engine.

This sounds like a good idea, padding packets seem specific to WebRTC, but probably not possible without refactoring - after inbound RTX has been added, the padding packets are sent with retransmission SSRC and RTXParser drops them (they cannot be parsed as they do not contain a payload with the original sequence number)

PR #143, opened 2023-02-08
Replace buffer store with a map

mickel8:
This implementation should be much more performant

blue and yellow metrics are multiplied by 1_000_000

Here is an example plot showing number of reductions when using BufferStore

visualization (1)

and here is a map
visualization (2)

While the x-axis differ a little, we can see that after 240s of talk we have ~37M of reductions vs 25M

In one case we still noticed some peaks for BufferStore. The overall number of reductions was also very high ~80M

visualization (3)

### DIFF
--- a/lib/membrane/rtp/outbound_rtx_controller.ex
+++ b/lib/membrane/rtp/outbound_rtx_controller.ex
@@ -19,7 +19,7 @@ defmodule Membrane.RTP.OutboundRtxController do
   @min_rtx_interval 10
 
   @impl true
-  def handle_init(_opts), do: {:ok, %{store: %{}, last_rtx_times: %{}}}
+  def handle_init(_opts), do: {:ok, %{store: %{}}}
 
   # Ignores padding packets
   # TODO: Should it?
--- a/lib/membrane/rtp/outbound_rtx_controller.ex
+++ b/lib/membrane/rtp/outbound_rtx_controller.ex
@@ -26,7 +26,7 @@ defmodule Membrane.RTP.OutboundRtxController do
   @impl true
   def handle_process(:input, buffer, _ctx, state) when byte_size(buffer.payload) > 0 do
     idx = rem(buffer.metadata.rtp.sequence_number, @max_store_size)
-    state = put_in(state, [:store, idx], {System.monotonic_time(:millisecond), buffer})
+    state = put_in(state, [:store, idx], {nil, buffer})
     {{:ok, forward: buffer}, state}
   end
 
@@ -47,13 +47,14 @@ defmodule Membrane.RTP.OutboundRtxController do
         maybe_retransmit(seq_num, now, store)
       end)
 
-    buffers_to_retransmit = Enum.reject(buffers, fn buffer -> buffer == nil end)
+    buffers_to_retransmit = Enum.reject(buffers, &is_nil/1)
 
-    unless buffers_to_retransmit == [],
-      do:
-        Membrane.Logger.info(
-          "Retransmitting buffers (#{Enum.count(buffers_to_retransmit)}) with the following sequence numbers:\n#{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}"
-        )
+    unless buffers_to_retransmit == [] do
+      Membrane.Logger.info("""
+      Retransmitting (#{length(buffers_to_retransmit)}) buffers, sequence numbers:
+      #{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}
+      """)
+    end
 
     {{:ok, buffer: {:output, buffers_to_retransmit}}, %{state | store: store}}
   end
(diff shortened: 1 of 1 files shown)


lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -51,49 +42,35 @@ defmodule Membrane.RTP.OutboundRtxController do
...
-    state =
-      Map.update!(state, :last_rtx_times, fn times ->
-        updates =
-          Map.new(buffers_to_retransmit, fn buffer ->
-            {buffer.metadata.rtp.sequence_number, now}
-          end)
-
-        Map.merge(times, updates)
-      end)
+    buffers_to_retransmit = Enum.reject(buffers, fn buffer -> buffer == nil end)
```
bblaszkow06, line 50:
```suggestion
    buffers_to_retransmit = Enum.reject(buffers, &is_nil/1)
```

lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -51,49 +42,35 @@ defmodule Membrane.RTP.OutboundRtxController do
...
-    if Enum.count(store) > @max_store_size do
-      {_entry, store} = BufferStore.flush_one(store)
-      maintain_store_size(store)
+  defp maybe_retransmit(seq_num, now, store) do
+    idx = rem(seq_num, @max_store_size)
+    {last_rtx_timestamp, buffer} = Map.get(store, idx, {nil, nil})
+
+    if buffer != nil and buffer.metadata.rtp.sequence_number == seq_num and
+         now - last_rtx_timestamp >= @min_rtx_interval do
+      store = Map.put(store, seq_num, {now, buffer})
```
bblaszkow06, line 70: IIUC, you should be putting the tuple under idx, not `seq_num`.
```suggestion
      store = Map.put(store, idx, {now, buffer})
```

lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -51,49 +42,35 @@ defmodule Membrane.RTP.OutboundRtxController do
...
-        Map.merge(times, updates)
-      end)
+    buffers_to_retransmit = Enum.reject(buffers, fn buffer -> buffer == nil end)
 
     unless buffers_to_retransmit == [],
       do:
         Membrane.Logger.info(
-          "Retransmitting buffers with the following sequence numbers:\n#{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}"
+          "Retransmitting buffers (#{Enum.count(buffers_to_retransmit)}) with the following sequence numbers:\n#{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}"
         )
```
bblaszkow06, line 56: I'd compact this log a bit: (cannot make suggestions due to deleted lines, GH sux)

```elixir
    unless buffers_to_retransmit == [] do
      Membrane.Logger.info("""
      Retransmitting (#{length(buffers_to_retransmit)}) buffers, sequence numbers:
      #{inspect(Enum.map(buffers_to_retransmit, & &1.metadata.rtp.sequence_number))}
      """)
    end
```

bblaszkow06 (changes requested):

lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -20,23 +19,15 @@ defmodule Membrane.RTP.OutboundRtxController do
...
-        {:ok, new_store} ->
-          maintain_store_size(new_store)
-
-        {:error, :late_packet} ->
-          store
-      end
-    end)
-    |> then(&{{:ok, forward: buffer}, &1})
+    idx = rem(buffer.metadata.rtp.sequence_number, @max_store_size)
+    state = put_in(state, [:store, idx], {System.monotonic_time(:millisecond), buffer})
```
bblaszkow06, line 29: One more thing: this changes the behaviour of the controller as we won't retransmit any packet within 10ms. I assume the receiver is using some JitterBuffer and will wait a bit before sending NACK to compensate for possible reorder, thus we should handle first NACK right away.

I even wonder if we should have any `min_rtx_interval` at all, as it seems large retransmissions are only a symptom, not a root cause of our problems

mickel8:
New figures 

In ~240 seconds we perform ~27-30M of reductions which is about 8-11M less than using BufferStore

`total_buffers_sent` is multiplied by 1000, awaiting buffers and message_queue_len are multiplied by 10 000 000

visualization (8)

visualization (7)

visualization (5)

lib/membrane/rtp/outbound_rtx_controller.ex
```diff
@@ -20,23 +19,15 @@ defmodule Membrane.RTP.OutboundRtxController do
   @min_rtx_interval 10
 
   @impl true
-  def handle_init(_opts), do: {:ok, %{store: BufferStore.new(), last_rtx_times: %{}}}
+  def handle_init(_opts), do: {:ok, %{store: %{}, last_rtx_times: %{}}}
```
bblaszkow06, line 22:
```suggestion
  def handle_init(_opts), do: {:ok, %{store: %{}}}
```

bblaszkow06 (approved):

PR #144, opened 2023-02-14
Auth Fail

daniel-jodlos:
It seems that sometimes we receive `:auth_fail` error from SRTP Decryptor.
This is concerning, but may be caused by a late packet after SDP renegotiation.
It may be useful to not instantly crash the endpoint in this scenario and verify if `:auth_fail`s are not constant.


bblaszkow06 (approved):
For the record - it doesn't seem to be related to renegotiation - I had that error in a 1on1 call with screen sharing. The call was left running while I attended meetings and logs indicate one peer crashed with auth_fail, then the room was active for some time until the other crashed as well

mickel8:
JFYI, I added issue to investigate this - RTC-144

PR #148, opened 2023-03-16
Mv 309 implement vad (from master branch)

pkrucz00:
A prototype of implementation of Voice Acivity Detection algorithm. The implementation is based on the algorithm proposed in  [*Dominant Speaker Identification for Multipoint
Videoconferencing*](https://israelcohen.com/wp-content/uploads/2018/05/IEEEI2012_Volfin.pdf). 

Changes in regard to the original include:
   - it is not a Dominant Speaker Identification algorithm but Voice Activity Detection which means:
         - there is no algorithm for picking the best activity score
         - the activity score is thresholded to estimate if the user is speaking
  - the *immediates* are thresholded by a simple dB level value so it's not adjusted to the browser's changes of the user's volume 

What I do like about this:
   - the algorithm is a lot more responsive than the previous algorithm based on the running average
   - it does not use any hard and heavy ML like neuron networks or transformers; it's just simple approximation of speech based on probability and threshold
   - it's pretty fast even without optimisations

What I do not like:
   - the module is not documented
   - there are no tests
   - parameters are not tuned well which makes the icon jump way too frequently

There is still a big room for improvement for this algorithm. I leave the POC of what has been accomplished so far to see if it is of any use. Hope you find it at least a little interesting.


lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  defp count_active(data, threshold) do
+    Enum.reduce(data, 0, fn x, acc -> acc + indicator_func(x, threshold) end)
+  end
+
+  defp compute_interval(littles, sub_list_length, threshold) do
+    littles
+    |> Enum.chunk_every(sub_list_length)
+    |> Enum.map(&count_active(&1, threshold))
```
bblaszkow06, line 61: Why do you map the values you won't be using? I wanted my initial suggestion to be
```suggestion
    littles
    |> Enum.chunk_every(sub_list_length)
    |> Enum.map(&count_active(&1, threshold))
    |> hd()
```
as you always take the head in `estimate_is_speaking` but this is equivalent to
```suggestion
    littles
    |> Enum.take(sub_list_length)
    |> count_active(threshold)
```
R U sure this is what you wanted to do?

lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
```diff
@@ -0,0 +1,118 @@
...
+  def trim_queue_and_estimate_vad(queue, threshold) do
+    trimmed_queue = safe_trim_queue(queue, @min_levels_length)
+
+    estimation =
+      trimmed_queue
+      |> Enum.map(fn {level, _timestamp} -> 127 + level end)
+      |> estimate_is_speaking(threshold)
+
+    {trimmed_queue, estimation}
+  end
```
bblaszkow06, line 117: This is a bit confusing when it comes to separating the responsibilities. Is the queue internal to the `IsSpeakingEstimator` module or not? If yes, then it should be responsible also for putting the audio levels inside (BTW it could then apply the offset as now you're mapping the entries on each call). Otherwise, it should not be responsible for trimming the queue.

lib/membrane/rtp/vad.ex
```diff
@@ -132,85 +101,37 @@ defmodule Membrane.RTP.VAD do
...
-
-  defp get_audio_levels_vad(state) do
-    if state.audio_levels_count >= state.min_packet_num and avg(state) >= state.vad_threshold,
-      do: :speech,
-      else: :silence
+    %{state | audio_levels: audio_levels}
   end
 
-  defp avg(state), do: state.audio_levels_sum / state.audio_levels_count
+  defp update_queue(new_queue, state), do: %{state | audio_levels: new_queue}
```
bblaszkow06, line 124: This helper is a bit of an overkill, plus it's not used everywhere you update the queue, I'd got rid of it

bblaszkow06 (changes requested):
This was just a quick glance as I haven't got into the paper details yet.

lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
pkrucz00: This function is used for computing immediates, mediums and longs. It's true only the head is used for evaluating the activity score of each of them...

 *but*
When the mediums are computed, all the values of immediates are used and when longs are computed all values of mediums are needed. Moreover, `longs` has at most 1 element. 
Therefore I think those changes do not apply here. We can think about optimising this process in future, but as for now I don't seethe merit in this either.

lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
```diff
@@ -0,0 +1,118 @@
...
+  @min_activity_score 1.0e-8
+
+  defp binomial_coefficient(n, k) when k < 0 or k > n, do: 0
+  defp binomial_coefficient(n, k) when k == 0 or n == k, do: 1
+
+  defp binomial_coefficient(n, k) do
+    k = min(k, n - k)
+
+    Enum.to_list(1..k)
+    |> Enum.reduce(1, fn i, acc -> div(acc * (n - i + 1), i) end)
```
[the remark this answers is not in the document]
pkrucz00: More about the binomial coefficient:
   - it is based on the implementation given on wikipedia [here](https://en.wikipedia.org/wiki/Binomial_coefficient#In_programming_languages) (the second code snippet with iterative implementation in python)
   - as noted in the conversation with @bblaszkow06, the values evaluated here are no bigger than `max(@n1, @n2, @n3)`. This means this could be evaluated in compile time.

lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
pkrucz00: I'll make `vad.ex` responsible for the queue and change the range of values inside of it from [-127, 0] to [0, 127]. Then I'll change the module of the given function.

pkrucz00:
What has been added since he last review:
   - unit tests
   - changing what has been pointed out in the recent review

bblaszkow06 (changes requested):

.gitignore
```diff
@@ -182,3 +182,4 @@ $RECYCLE.BIN/
 # End of https://www.gitignore.io/api/c,vim,linux,macos,elixir,windows,visualstudiocode
 
 !test/fixtures/**/*.hex
+.vscode
```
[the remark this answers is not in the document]
pkrucz00: I've overlooked that `.gitignore` has an exception for `settings.json` file where I'd put some words to the vscode dictionary. I thought it was negligible. 
But if the rule was in the template then I think it was for there for some good reason. I'll add the `settings.json` file and delete this line.

lib/membrane/rtp/vad_utils/is_speaking_estimator.ex
```diff
@@ -0,0 +1,109 @@
...
+    do: :silence
+
+  def estimate_is_speaking(levels, level_threshold) do
+    immediates = compute_immediates(levels, level_threshold)
+    mediums = compute_mediums(immediates)
+    longs = compute_longs(mediums)
+
+    immediate_score = immediates |> hd() |> compute_activity_score(@n1, 1)
+    medium_score = mediums |> hd() |> compute_activity_score(@n2, 24)
+    long_score = longs |> hd() |> compute_activity_score(@n3, 47)
```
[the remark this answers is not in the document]
pkrucz00: Those numbers are lambda parameters in the exponential distribution. It's algorithm specific, but I think it could be added as a parameter if someone wanted to tweak it jus a bit. 

I didn't want to add it as another parameter because there is already 8 of them. I think it is a lot but can't figure out if the number can be reduced.

pkrucz00 (commented):
I've applied most of the comments to the code. One is left for the discussion. Hope it looks better now:)

lib/membrane/rtp/vad.ex
```diff
@@ -130,87 +98,34 @@ defmodule Membrane.RTP.VAD do
...
-            audio_levels_count: state.audio_levels_count - 1,
-            audio_levels: audio_levels
-        }
-
-        {:cont, state}
-      else
-        {:halt, state}
-      end
-    end)
-  end
```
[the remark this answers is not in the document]
bblaszkow06: OK, you're right it handles out-of-order scenarios. It will still be broken for missing packets as you're not putting anything in the queue and it might happen that `IsSpeakingEstimator` will receive much more than the desired time interval since it's based on queue length.

I guess though unless it's trivial, it can be done in separate task/issue

lib/membrane/rtp/vad.ex
```diff
@@ -4,16 +4,9 @@ defmodule Membrane.RTP.VAD do
...
-    - mapping them to db (range [0, 127])
-    - calculating the epoch of the timestamp
-    - storing them in a queue
-    - asking the IsSpeakingEstimator if the audio levels indicate speech or silence
-    - emitting the VAD event if the estimation has changed
+  Sends `Membrane.RTP.VadEvent` when a score from `Membrane.RTP.VadUtils.IsSpeakingEstimator` changes.
 
-
-  The more detailed explanation of how the VAD algorithm can be found in the IsSpeakingEstimator module
+  A more detailed explanation of how the VAD algorithm can be found in the `Membrane.RTP.VadUtils.IsSpeakingEstimator` module
```
bblaszkow06, line 9: The module name has been changed
```suggestion
  A more detailed explanation of how the VAD algorithm can be found in the `Membrane.RTP.Vad.IsSpeakingEstimator` module
```

lib/membrane/rtp/vad/is_speaking_estimator.ex
```diff
@@ -1,70 +1,64 @@
...
-    - subunit_threshold:  number equal or above which the number of active subunits must be
-                          for the given tier to be marked as active,
-                          (for immediates it is equal to the threshold given as a estimate_is_speaking/2 argument)
-    - lambda:             parameter for the exponential distribution (element of the activity score computations)
+  | Parameter | Description |
+  |-|-|
+  | `subunits` | number of smaller units in one bigger unit |
+  | `score_threshold`  | number equal or above which the activity score of the given tier must be to be counted as indicating speech |
+  | `subunit_threshold` | number equal or above which the number of active subunits must be for the given tier to be marked as active (for immediates it is equal to the threshold given as a estimate_is_speaking/2 argument) |
+  | `lambda` | parameter for the exponential distribution (element of the activity score computations) |
```
bblaszkow06, line 57: This still doesn't show how to configure the params - a code snippet to put in config should handle that

lib/membrane/rtp/vad.ex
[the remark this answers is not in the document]
pkrucz00: The previous algorithm didn't handle it. Only a `rtp_timestamp_increment` was measured (a quotient of time between packets and the clock_rate).
My idea for implementing it is:
  1. Measuring `rtp_timestamp_increment`.
  2. Computing the difference between new timestamp and previous timestamp.
  3. Checking how many timestamps there should be by dividing the difference by `rtp_timestamp_increment`.
  4. Padding the missing packets with an audio level set to `0`.

For example: if `rtp_timestamp_increment` is 960 and the difference between `current` and `previous` rtp timestamp is 1920, we add to the `audio_levels` queue:
  - one `0` for the missing packet and the current audio level value,
  - the value of audio level in the rtp packet that have arrived


This is a simple solution I have come up with. It solves one problem you have proposed. Is this enough?

lib/membrane/rtp/vad/is_speaking_estimator.ex
```diff
@@ -0,0 +1,179 @@
...
+  and based on a threshold given as a second input
+  computes if the person is speaking.
+
+  The input levels are interpreted on 3 tiers. Each tier consists of intervals specified below:
+
+  | Name         | Interpretation       | Number of RTP packets (default) | length    |
+  |--------------|----------------------|---------------------------------|-----------|
+  | `immediate`  | smallest sound chunk | 1                               | ~20 [ms]  |
+  | `medium`     | one word             | 1 \* 10 = 10                    | ~200 [ms] |
+  | `long`       | half/one sentence    | 1 \* 10 \* 7 = 70               | ~1.4 [s]  |
```
bblaszkow06, line 19: This line is rendered incorrectly, although it seems like a bug in `ex_doc`
. Here's the fix:
```suggestion
  | `long`       | half/one sentence    | 1 \\* 10 \* 7 = 70              | ~1.4 [s]  |
```

You should probably also file a bug at [`ex_doc`](https://github.com/elixir-lang/ex_doc/issues) repo (it is possible the issue is deeper, in one the ex_doc deps but I'd start from the `ex_doc` team) with a minimal example of reproduction.

lib/membrane/rtp/vad/is_speaking_estimator.ex
```diff
@@ -0,0 +1,179 @@
...
+  > **Example**
+  >
+  > If `level_threshold` is `90`, levels are `[80, 90, 100, 90]` and there are 2 levels in a immediate
+  > then `immediates` would be equal to `[1, 2]` since subunit of `[80, 90]` has 1 item above or equal to
+  > the threshold and subunit `[100, 90]` has 2 such items.
+  >
+  > Same goes for mediums. If `medium subunit threshold` is 2 and number of subunits is 2
+  > then `mediums` are equal to `[1]` since the only subunit `[1, 2]` had only one element above or equal to the threshold.
+
+  The number of levels the function requires is equal to the product of all subunits per tier.
```
bblaszkow06, line 36:
```suggestion
  The number of levels the function requires equals the product of the subunits number required for each tier.
```

bblaszkow06 (dismissed):
A few more details to fix in the docs and you can merge the PR.
approval-4

bblaszkow06 (approved):

PR #154, opened 2023-09-12
Remove RTP extensions on removing SessionBin's output pad

mickel8:
Looks like we haven't been removing RTP extensions so far... :thinking:


lib/membrane/rtp/session_bin.ex
```diff
@@ -609,17 +609,14 @@ defmodule Membrane.RTP.SessionBin do
...
-        child = Map.get(ctx.children, name)
-        child && not child.terminating?
+      ctx.children
+      |> Enum.filter(fn {name, child} ->
+        # we use pattern matching to match all
+        # extensions spawned for given ssrc stream
+        # e.g. vad
+        match?({_name, ^ssrc}, name) && not child.terminating?
       end)
+      |> Enum.map(fn {name, _child} -> name end)
```
bblaszkow06, line 619: nitpick: I'd probably use the pattern in function head and maybe `flat_map` to save one iteration (assuming `terminating?` cannot be `nil`)
```suggestion
      ctx.children
      |> Enum.flat_map(fn 
        {{name, ^ssrc}, %{terminating?: false}} -> [name]
        _other -> []
      end)
```

bblaszkow06 (dismissed):

lib/membrane/rtp/session_bin.ex
FelonEkonom, line 619: Once `name` is whole name of the child (eg. `{:vad, 123456}`), another time `_name` is just type of the extension (eg. `:vad`) - IMO it would be cleaner to not use the same variable name for different things

FelonEkonom (approved):

bblaszkow06 (dismissed):

lib/membrane/rtp/session_bin.ex
```diff
@@ -609,14 +609,10 @@ defmodule Membrane.RTP.SessionBin do
...
 
     to_remove =
-      ctx.children
-      |> Enum.filter(fn {name, child} ->
-        # we use pattern matching to match all
-        # extensions spawned for given ssrc stream
-        # e.g. vad
-        match?({_name, ^ssrc}, name) && not child.terminating?
+      Enum.flat_map(ctx.children, fn
+        {{name, ^ssrc}, %{terminating?: false}} -> [name]
```
FelonEkonom, line 613:
```suggestion
        {{name, ^ssrc}, %{terminating?: false}} -> [{name, ssrc}]
```

PR #160, opened 2023-11-27
Fix remove_children and end_of_stream

mat-hek:


mickel8 (approved):
Brilliant, thank you! :bulb:

PR #161, opened 2024-01-10
Add tcp depayloader

Noarkhh:
https://github.com/membraneframework/membrane_core/issues/697
Add an element for depayloading a RTP stream received through TCP and encapsulated according to RFC 7826 Section 14.


lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+                Channel identifier which encapsulated RTP packets will have.
+                """
+              ],
+              rtsp_session: [
+                spec: pid() | nil,
+                default: nil,
+                description: """
+                PID of a RTSP Session (returned from Membrane.RTSP.start or Membrane.RTSP.start_link)
+                that received RTSP responses will be forwarded to. If nil the responses will be
+                discarded.
```
varsill, line 23:
```suggestion
                PID of an RTSP Session (returned from `Membrane.RTSP.start` or `Membrane.RTSP.start_link`)
                to which the received RTSP responses will be forwarded. If `nil`, the responses will be
                discarded.
```

lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+      })
+
+    {[], state}
+  end
+
+  @impl true
+  def handle_stream_format(:input, _stream_format, _ctx, state) do
+    stream_format = %RemoteStream{type: :packetized, content_format: RTP}
+    {[stream_format: {:output, stream_format}], state}
+  end
```
varsill, line 46: Are you sure we want to send `stream_format` each time `handle_stream_format` callback is invoked? Perhaps it would be better do send it in `handle_playing`?

lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  @impl true
+  def handle_buffer(
+        :input,
+        %Buffer{payload: payload, metadata: metadata},
+        _ctx,
+        %{unprocessed_data: unprocessed_data, rtsp_session: rtsp_session} = state
```
varsill, line 53: [NIT] Wouldn't it be not to match on field values and instead refer to `state.unprocessed_data` in the function body?

lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+      if rtsp_response?(unprocessed_data, payload) do
+        if rtsp_session != nil do
+          {:ok, %RTSP.Response{status: 200}} =
+            RTSP.handle_response(rtsp_session, unprocessed_data)
+        end
+
+        <<>>
+      else
+        unprocessed_data
+      end
```
varsill, line 65: [NIT] You can consider using `cond` instead of nested if, but do it as you like :p

lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+
+    packets_binary = unprocessed_data <> payload
+
+    {unprocessed_data, complete_packets_binaries} =
+      get_complete_packets(packets_binary, state.rtp_channel_id)
+
+    packets_buffers =
+      Enum.map(complete_packets_binaries, &%Buffer{payload: &1, metadata: metadata})
```
varsill, line 73: do we have a guarantee that the `metadata` is rewritten properly? What if the buffer's payload is put in `unprocessed_data`, and the metadata is attached to the data corresponding to the previous input buffer?

lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -0,0 +1,119 @@
...
+
+    {[buffer: {:output, packets_buffers}], %{state | unprocessed_data: unprocessed_data}}
+  end
+
+  @spec rtsp_response?(binary(), binary()) :: boolean()
+  defp rtsp_response?(maybe_rtsp_response, new_payload) do
+    # If new payload starts with a "$" and unprocessed data started with "RTSP" then we know
+    # the unprocessed data should be a RTSP message
+    String.starts_with?(new_payload, "$") and String.starts_with?(maybe_rtsp_response, "RTSP")
```
varsill, line 82: I think this piece of code is self explanatory - I would rather have a brief explanation what `$` means :D

varsill (changes requested):

lib/membrane/rtp/tcp_depayloader.ex
Noarkhh:
```
cond do
  not rtsp_response?(state.unprocessed_data, payload) ->
    state.unprocessed_data

  state.rtsp_session != nil ->
    {:ok, %RTSP.Response{status: 200}} =
      RTSP.handle_response(state.rtsp_session, state.unprocessed_data)
    <<>>

  state.rtsp_session == nil ->
    <<>>
end
``` 
Something like this seems reasonable, but the fact that `<<>>` is always returned if `rtsp_response?` returns true makes me prefer the nested if

lib/membrane/rtp/tcp_depayloader.ex
Noarkhh: the only metadata that is in buffers coming from TCP Source are `tcp_source_address`, `tcp_source_port` and `arrival_ts`. The first two don't change and the third one can be interpreted as in when the packet has fully arrived, which then works in this case

varsill: I see, it makes sense, however, I am slightly worried that in case we were about to extend the `metadata` sent from the `TCP Source`, we would need to remember that the metadata is translated by one buffer when passed through the depayloader

varsill (dismissed):
I've left one comment unresolved where I have described my doubts, but generally speaking - LGTM ;)

mat-hek (approved):

PR #162, opened 2024-02-21
Rename RTP.TCP.Depayloader to RTP.RTSP.Decapsulator

Noarkhh:


lib/membrane/rtp/tcp_depayloader.ex
```diff
@@ -1,7 +1,8 @@
-defmodule Membrane.RTP.TCP.Depayloader do
+defmodule Membrane.RTP.RTSP.Depayloader do
```
wkozyra95, line 1: PR name does not match the change

lib/membrane/rtp/rtsp_decapsulator.ex
```diff
@@ -95,12 +96,10 @@ defmodule Membrane.RTP.TCP.Depayloader do
...
 
   defp get_complete_packets(
-         <<"$", _rest::binary>> = packets_binary,
+         <<"$", received_channel_id, payload_length::size(16), rest::binary>> = packets_binary,
          channel_id,
          complete_packets
        ) do
-    <<"$", received_channel_id, payload_length::size(16), rest::binary>> = packets_binary
-
     if payload_length > byte_size(rest) do
```
mat-hek, line 103: This would be more idiomatic

```elixir
case rest do
   <<complete_packet_binary::binary-size(payload_length)-unit(8), rest::binary>> -> ...
```

mat-hek (dismissed):

mat-hek (approved):

PR #163, opened 2024-02-22
Implement elements for packetizing and depacketzing RTP packets for TCP

wkozyra95:
- `Membrane.RTP.TCP.Encapsulator` - reads packet and generates bytestream that can be send over  connection-oriented transport (such as TCP)
- `Membrane.RTP.TCP.Decapsulator` - create RTP packets from a bytestream

Mostly based on `Membrane.RTP.TCP.Depayloader` (this element implements different mechanism used in RTSP)

I'm open to the name suggestions:
- I decided to go with packetizer/depacketizer so as not to confuse it with payloaders/depayloared for RTP payload like h264
- Other alternative could be to name it `Membrane.RTP.TCP.Depayloader` and `Membrane.RTP.TCP.Payloader`, but in that case existing `Membrane.RTP.TCP.Depayloader` would need to be renamed

### DIFF
--- a/lib/membrane/rtp/tcp_depacketizer.ex
+++ b/lib/membrane/rtp/tcp_depacketizer.ex
@@ -1,0 +1,48 @@
+defmodule Membrane.RTP.TCP.Depacketizer do
+  @moduledoc """
+  This element provides functionality of serializing RTP and RTCP packets into a bytestream
+  that can be send over TCP connection. The encapsulation is described in RFC 4571.
+
+  Packets in the stream will have the following structure:
+  [Length :: 2 bytes][packet :: <Length> bytes]
+  """
+  use Membrane.Filter
+
+  alias Membrane.{Buffer, RemoteStream, RTP}
+
+  def_input_pad :input, accepted_format: %RemoteStream{type: :packetized, content_format: RTP}
+
+  def_output_pad :output, accepted_format: %RemoteStream{type: :bytestream}
+
+  @impl true
+  def handle_init(_ctx, _opts) do
+    {[], %{}}
+  end
+
+  @impl true
+  def handle_playing(_ctx, state) do
+    stream_format = %RemoteStream{type: :bytestream}
+    {[stream_format: {:output, stream_format}], state}
+  end
+
+  @impl true
+  def handle_stream_format(:input, _stream_format, _ctx, state) do
+    {[], state}
+  end
+
+  @impl true
+  def handle_buffer(:input, %Buffer{payload: payload, metadata: metadata}, _ctx, state) do
+    len_bytes =
+      case :binary.encode_unsigned(byte_size(payload), :big) do
+        <<len::size(8)>> -> <<0, len>>
+        <<len::binary-size(2)>> -> len
+      end
+
+    buffer = %Buffer{
+      payload: len_bytes <> payload,
+      metadata: metadata
+    }
+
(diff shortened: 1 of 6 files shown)


lib/membrane/rtp/tcp_depacketizer.ex
```diff
@@ -0,0 +1,48 @@
...
+    {[], state}
+  end
+
+  @impl true
+  def handle_buffer(:input, %Buffer{payload: payload, metadata: metadata}, _ctx, state) do
+    len_bytes =
+      case :binary.encode_unsigned(byte_size(payload), :big) do
+        <<len::size(8)>> -> <<0, len>>
+        <<len::binary-size(2)>> -> len
+      end
```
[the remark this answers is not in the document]
Noarkhh: this can be further expanded to `payload = <<byte_size(payload)::size(16), payload::binary>>`, which creates the prefixed payload that can be directly passed to outputed buffer

lib/membrane/rtp/tcp_depacketizer.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Membrane.RTP.TCP.Depacketizer do
```
Noarkhh, line 1: It was kind of counterintuitive to me that Depacketizer is reponsible for encapsulation, even though it technically is correct. What do you think about Encapsulator and Decapsulator?

lib/membrane/rtp/tcp_packetizer.ex
```diff
@@ -0,0 +1,64 @@
...
+          {unprocessed_data :: binary(), complete_packets :: [binary()]}
+  defp get_complete_packets(packets_binary, complete_packets \\ [])
+
+  defp get_complete_packets(packets_binary, complete_packets)
+       when byte_size(packets_binary) <= 2 do
+    {packets_binary, Enum.reverse(complete_packets)}
+  end
+
+  defp get_complete_packets(packets_binary, complete_packets) do
+    <<payload_length::size(16), rest::binary>> = packets_binary
```
mat-hek, line 55: you can change the function headers to the following:
```elixir
defp get_complete_packets(
  <<payload_length::16, payload::binary-size(payload_length), rest::binary>>,
  complete_packets) do
# next chunk available
end
defp get_complete_packets(leftover, complete_packets) do
# next chunk unavailable
end
```

lib/membrane/rtp/tcp_depacketizer.ex
wkozyra95: Personally, I would probably stay with Packetizer/Depacketizer or Payloader/Depayloader, but I don't mind changing that

mat-hek: Hmm, actually is it correct? I think of depacketizing as of getting out of packets, which in this case would be removing the two-byte header with packet length…

mat-hek: the RFC calls it framing: https://datatracker.ietf.org/doc/html/rfc4571 however as I read it, I understand that by framing they mean adding the 2 byte header, not the other way

wkozyra95: packetizing converting to packets from a continuous stream of data (TCP does not really have a concept of packets)

mat-hek (approved):

Noarkhh (dismissed):
Remember to bump version in mix.exs and readme

PR #170, opened 2024-10-29
Skip empty packets

mat-hek:
WebRTC sends empty RTP packets for probing bandwidth. If they're not skipped, they can break further processing.


varsill (dismissed):
Just remember to bump version in README ;)

varsill (approved):

PR #171, opened 2024-11-13
RTP Demuxer

Noarkhh:
closes https://github.com/membraneframework/membrane_core/issues/884

RTP Demuxer has a single static input pad and dynamic output pads. It receives a raw RTP stream on the input pad and demuxes it onto the output pads based on SSRCs. Goal of introducing this element is to ultimately provide a viable alternative to the really complex RTP.SessionBin.


mat-hek (changes requested):

lib/membrane/rtp/demuxer.ex
```diff
@@ -0,0 +1,170 @@
+defmodule Membrane.RTP.Demuxer do
+  @moduledoc """
+  Element capable of receiving a raw RTP stream and demuxing it into individual parsed streams based on packet ssrcs. 
+  Whenever a new stream is recognized, a `new_rtp_stream_notification` is sent to the element's parent. In turn it should link a 
+  `Membrane.Pad.ref({:output, ssrc})` output pad of this element to receive the stream specified in the notification.
```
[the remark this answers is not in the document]
Noarkhh: Agree, that would be nice. Did you have any ideas what would be a good way to match these pads connected upfront to rtp tracks? Have pad options for encoding and payload type and base on them?

lib/membrane/rtp/demuxer.ex
```diff
@@ -67,25 +118,78 @@ defmodule Membrane.RTP.Demuxer do
...
-    {[stream_format: {pad, %RTP{}}] ++ Enum.reverse(buffered_actions), state}
+    stream_state = state.stream_states[matching_stream_ssrc]
+
+    case stream_state do
+      nil ->
+        state = %{
+          state
+          | pads_waiting_for_stream:
+              Map.put(state.pads_waiting_for_stream, pad, ctx.pad_options.stream_id)
+        }
```
mat-hek, line 137:
```suggestion
        state = put_in(state.pads_waiting_for_stream[pad], ctx.pad_options.stream_id)
```

lib/membrane/rtp/demuxer.ex
```diff
@@ -67,25 +118,78 @@ defmodule Membrane.RTP.Demuxer do
...
+      %{phase: :waiting_for_link} ->
+        Process.cancel_timer(stream_state.link_timer)
+
+        buffer_action = [buffer: {pad, Enum.reverse(stream_state.queued_buffers)}]
+
+        end_of_stream_action =
+          if stream_state.end_of_stream_buffered, do: [end_of_stream: pad], else: []
+
+        state =
+          Bunch.Struct.put_in(
```
mat-hek, line 150: Generally use `Kernel.put_in` instead of `Bunch.Struct.put_in`

lib/membrane/rtp/demuxer.ex
```diff
@@ -109,61 +214,170 @@ defmodule Membrane.RTP.Demuxer do
...
+    %{clock_rate: clock_rate} =
+      RTP.PayloadFormat.resolve(
+        payload_type: packet.payload_type,
+        payload_type_mapping: payload_type_mapping
+      )
+
+    pts =
+      if clock_rate == nil,
+        do: nil,
+        else: Ratio.new(packet.timestamp, clock_rate) |> Membrane.Time.seconds()
```
mat-hek, line 376: Can we cache the clock rate to avoid resolving it for every buffer?

lib/membrane/rtp/demuxer.ex
```diff
@@ -32,25 +29,79 @@ defmodule Membrane.RTP.Demuxer do
...
+      ]
+    ]
+
+  def_options payload_type_mapping: [
+                spec: RTP.PayloadFormat.payload_type_mapping(),
+                default: %{},
+                description: "Mapping of the custom RTP payload types ( > 95)."
+              ],
+              not_linked_pad_handling: [
+                spec: %{action: :raise | :ignore, timeout: Membrane.Time.t()},
```
mat-hek, line 61: could we also have a 'warn' action?

lib/membrane/rtp/demuxer.ex
```diff
@@ -0,0 +1,384 @@
...
+  """
+  @type output_metadata :: %{rtp: ExRTP.Packet.t()}
+
+  @typedoc """
+  Notification sent by this element to it's parent when a new stream is received. Receiving a packet 
+  with previously unseen ssrc is treated as receiving a new stream.
+  """
+  @type new_rtp_stream_notification ::
+          {:new_rtp_stream, ssrc :: RTP.ssrc(), payload_type :: RTP.payload_type(),
+           extensions :: [ExRTP.Packet.Extension.t()]}
```
mat-hek, line 30: Let's make this a map or struct, so we can add more fields in the future

lib/membrane/rtp/demuxer.ex
```diff
@@ -0,0 +1,384 @@
...
+      )
+
+    pts =
+      if clock_rate == nil,
+        do: nil,
+        else: Ratio.new(packet.timestamp, clock_rate) |> Membrane.Time.seconds()
+
+    %Membrane.Buffer{
+      payload: packet.payload,
+      pts: pts,
```
mat-hek, line 380: To know the PTS, we'd need to handle timestamp rollovers. That's in the jitter buffer. Here we can only set the RTP timestamp in metadata

lib/membrane/rtp/demuxer.ex
Noarkhh: I do, but in cases, where keys are not known in compile time, I find Bunch.Struct functions handy

mat-hek: I think I don't see such cases in this PR 🤔

mat-hek (approved):

PR #173, opened 2024-12-03
RTP Muxer

Noarkhh:
closes https://github.com/membraneframework/membrane_core/issues/911


lib/membrane/rtp/muxer.ex
```diff
@@ -0,0 +1,179 @@
+defmodule Membrane.RTP.Muxer do
+  @moduledoc false
```
mat-hek, line 2: 🤔

test/membrane/rtp/demuxer_muxer_integration_test.exs
mat-hek, line 1: Can we have a test where we demux and mux packets two times and check if we get the same results? That would need to happen on a stream that's already past the jitter buffer.

lib/membrane/rtp/muxer.ex
```diff
@@ -0,0 +1,179 @@
...
+      buffer.pts
+      |> Membrane.Time.as_seconds()
+      |> Numbers.mult(stream_state.clock_rate)
+      |> Ratio.trunc()
+
+    timestamp = rem(stream_state.initial_timestamp + rtp_offset, @max_timestamp + 1)
+    sequence_number = rem(stream_state.sequence_number + 1, @max_sequence_number + 1)
+
+    state =
+      Bunch.Struct.put_in(state, [:stream_states, pad_ref, :sequence_number], sequence_number)
```
mat-hek, line 129: don't use Bunch.Struct

lib/membrane/rtp/muxer.ex
Noarkhh: I guess I forgor

lib/membrane/rtp/muxer.ex
```diff
@@ -1,5 +1,9 @@
 defmodule Membrane.RTP.Muxer do
-  @moduledoc false
+  @moduledoc """
+  Element that combines multiple streams into a single RTP stream. Each new input stream is assigned a unique SSRC, that the packets
```
mat-hek, line 3: The first line is used as a short description

```suggestion
  Element that combines multiple streams into a single RTP stream.
  
  Each new input stream is assigned a unique SSRC, that the packets 
```

test/membrane/rtp/demuxer_muxer_integration_test.exs
```diff
@@ -0,0 +1,125 @@
...
+        List.first(sorted_packets)
+
+      normalized_packets =
+        sorted_packets
+        |> Enum.map(fn packet ->
+          packet
+          |> Bunch.Struct.update_in(:ssrc, &(&1 - ssrc))
+          |> Bunch.Struct.update_in(:sequence_number, &(&1 - first_sequence_number))
+          # round to ignore insignificant differences in timestamps
+          |> Bunch.Struct.update_in(:timestamp, &round((&1 - first_timestamp) / 10))
```
mat-hek, line 119: Bunch.Struct 😡

mat-hek (dismissed):

PR #175, opened 2025-03-05
Demuxer jitter buffer

Noarkhh:
closes https://github.com/membraneframework/membrane_core/issues/915


mat-hek (commented):
As we spoke, let's try to move the jitter buffer functionality to a separate module

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+          pad_options() | nil,
+          RTP.PayloadFormat.payload_type_mapping()
+        ) :: State.t()
+  def initialize_state_with_new_rtp_stream(packet, pad_options, payload_type_mapping) do
+    {clock_rate, latency, initial_latency_waiting} =
+      case pad_options do
+        nil ->
+          {nil, nil, true}
+
+        %{use_jitter_buffer: false} = pad_options ->
```
mat-hek, line 61: as we spoke, let's use `latency: 0` for that

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+                  max_latency_timer: nil
+                ]
+  end
+
+  @spec initialize_state_with_new_rtp_stream(
+          ExRTP.Packet.t(),
+          pad_options() | nil,
+          RTP.PayloadFormat.payload_type_mapping()
+        ) :: State.t()
+  def initialize_state_with_new_rtp_stream(packet, pad_options, payload_type_mapping) do
```
mat-hek, line 55: how about renaming it to `new`, not passing `pad_options` at all, and requiring to always call `new` and then `complete_state_initialization_with_pad_options`?

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+  end
+
+  @spec complete_state_initialization_with_pad_options(
+          State.t(),
+          pad_options(),
+          RTP.payload_type(),
+          RTP.ssrc(),
+          RTP.PayloadFormat.payload_type_mapping()
+        ) :: State.t()
+  def complete_state_initialization_with_pad_options(
```
mat-hek, line 105:
```suggestion
  def initialize(
```

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+          {:error, :late_packet} ->
+            Membrane.Logger.debug("Late packet has arrived")
+            jitter_buffer_state
+        end
+    end
+  end
+
+  @spec get_actions(State.t(), Demuxer.stream_phase(), RTP.ssrc(), Membrane.Pad.ref()) ::
+          {[Membrane.Element.Action.t()], State.t()}
+  def get_actions(jitter_buffer_state, stream_phase, ssrc, pad) do
```
mat-hek, line 162: This name is quite generic, maybe something like
```suggestion
  def take_output(jitter_buffer_state, stream_phase, ssrc, pad) do
```

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+      jitter_buffer_state = set_timer(jitter_buffer_state, ssrc)
+
+      {actions, jitter_buffer_state}
+    else
+      {[], jitter_buffer_state}
+    end
+  end
+
+  @spec get_end_of_stream_actions(State.t(), Membrane.Pad.ref()) :: [Membrane.Element.Action.t()]
+  def get_end_of_stream_actions(jitter_buffer_state, pad) do
```
mat-hek, line 186:
```suggestion
  def finish(jitter_buffer_state, pad) do
```

lib/membrane/rtp/demuxer.ex
```diff
@@ -179,19 +218,54 @@ defmodule Membrane.RTP.Demuxer do
...
+    stream_state = state.stream_states[ssrc]
+
+    {actions, jitter_buffer_state} =
+      stream_state.jitter_buffer_state
+      |> JitterBuffer.get_actions(stream_state.phase, ssrc, stream_state.pad)
+
+    state =
+      put_in(
+        state.stream_states[ssrc].jitter_buffer_state,
+        %JitterBuffer.State{jitter_buffer_state | initial_latency_waiting: false}
```
mat-hek, line 235: I'd make it a function in the JitterBuffer

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+          Time.monotonic_time() - jitter_buffer_state.initialization_time
+
+        initial_latency_left = pad_options.jitter_buffer_latency - time_since_initialization
+
+        if initial_latency_left > 0 do
+          Process.send_after(
+            self(),
+            {:initial_latency_passed, ssrc},
+            Membrane.Time.as_milliseconds(initial_latency_left, :round)
+          )
```
mat-hek, line 131: shouldn't we set `initial_latency_waiting` accordingly?

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,262 @@
...
+          RTP.payload_type(),
+          RTP.ssrc(),
+          RTP.PayloadFormat.payload_type_mapping()
+        ) :: State.t()
+  def complete_state_initialization_with_pad_options(
+        jitter_buffer_state,
+        pad_options,
+        payload_type,
+        ssrc,
+        payload_type_mapping
```
mat-hek, line 110: We already pass these to `initialize_state_with_new_rtp_stream`, can we keep them in state not to pass them again?

mat-hek (changes requested):

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,233 @@
...
+    %State{jitter_buffer_state | max_latency_timer: nil}
+  end
+
+  @spec initial_latency_passed(State.t()) :: State.t()
+  def initial_latency_passed(jitter_buffer_state) do
+    %State{jitter_buffer_state | initial_latency_waiting: false}
+  end
+
+  @spec insert_buffer(State.t(), Membrane.Buffer.t(), Demuxer.stream_phase()) :: State.t()
+  def insert_buffer(jitter_buffer_state, buffer, stream_phase) do
```
varsill, line 114: I am not convinced we should pass `stream_phase` here - it introduces some "dependency loop" (`Demuxer` refers to `JitterBuffer` and the `JitterBuffer` refers to `Demuxer`. Perhaps that `case` clause could be part of the Demuxer's code?

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,233 @@
...
+          {:error, :late_packet} ->
+            Membrane.Logger.debug("Late packet has arrived")
+            jitter_buffer_state
+        end
+    end
+  end
+
+  @spec get_output_actions(State.t(), Demuxer.stream_phase()) ::
+          {[Membrane.Element.Action.t()], State.t()}
+  def get_output_actions(jitter_buffer_state, stream_phase) do
```
varsill, line 133: same as above

lib/membrane/rtp/demuxer/jitter_buffer.ex
```diff
@@ -0,0 +1,233 @@
...
+
+        buffer_ts ->
+          since_insertion = Time.monotonic_time() - buffer_ts
+          send_after_time = Time.as_milliseconds(latency - since_insertion, :round)
+
+          if send_after_time > 0 do
+            Process.send_after(self(), {:send_actions, jitter_buffer_state.ssrc}, send_after_time)
+          else
+            nil
+          end
```
varsill, line 186: [NIT]
```suggestion
          if send_after_time > 0, do: Process.send_after(self(), {:send_actions, jitter_buffer_state.ssrc}, send_after_time)
```

test/membrane/rtp/demuxer_muxer_integration_test.exs
```diff
@@ -10,6 +10,9 @@ defmodule Membrane.RTP.DemuxerMuxerTest do
     packets: 862
   }
 
+  @max_sequence_number Bitwise.bsl(1, 16) - 1
+  @max_timestamp Bitwise.bsl(1, 32) - 1
```
varsill, line 14: How about making `JitterBuffer`'s `@max_timestamp` public and referring it here?

test/membrane/rtp/demuxer_test.exs
```diff
@@ -60,6 +109,7 @@ defmodule Membrane.RTP.DemuxerTest do
     def handle_init(_ctx, opts) do
       spec =
         child(:pcap_source, %Membrane.Pcap.Source{path: opts.pcap_path})
+        |> child(if opts.reorder_packets, do: Reorderer, else: Membrane.Debug.Filter)
```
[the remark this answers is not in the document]
mat-hek: We actually now have the fake sink in the core 🎉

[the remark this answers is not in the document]
Noarkhh: But that's a filter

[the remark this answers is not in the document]
mat-hek: Ok, haven't noticed. Then let's do sth like
```suggestion
        |> then(if opts.reorder_packets, do: child(&1, Reorderer), else: & &1)
```

test/membrane/rtp/muxer_demuxer_integration_test.exs
```diff
@@ -24,8 +21,9 @@ defmodule Membrane.RTP.MuxerDemuxerTest do
         |> via_in(:input, options: [encoding: :H264])
         |> child(:rtp_muxer, Membrane.RTP.Muxer)
         |> child(:rtp_demuxer, Membrane.RTP.Demuxer)
-        |> via_out(:output, options: [stream_id: {:encoding_name, :H264}])
-        |> child(:jitter_buffer, %Membrane.RTP.JitterBuffer{clock_rate: clock_rate})
+        |> via_out(:output,
+          options: [stream_id: {:encoding_name, :H264}, jitter_buffer_latency: 0]
```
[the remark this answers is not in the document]
Noarkhh: I did that becaue all of the packets arrived before the initial latency passed, so when it finally did pass all of the packets were sent in a single batch, which I didn't like. I think I'll use Realtimer, I believe it makes sense in the context of Real-time Protocol

test/membrane/rtp/demuxer_muxer_integration_test.exs
varsill: I think `max_timestamp` could be a part of JitterBuffer's API, it might be useful for the user to be able to figure out what that `max_timestamp` is

lib/membrane/rtp/demuxer/jitter_buffer.ex
varsill: It feels slightly bloaty and that is why I thought of writing it in a more concise way, but sure, do what you are comfortable with doing ;)

lib/membrane/rtp/demuxer.ex
```diff
@@ -309,55 +321,87 @@ defmodule Membrane.RTP.Demuxer do
...
+          stream_state,
+          state.payload_type_mapping
+        )
 
-    {[notify_parent: create_new_stream_notification(packet)], state}
+      {_stream_id, state} = pop_in(state.pads_waiting_for_stream[pad_waiting_for_stream])
+      state = put_in(state.stream_states[packet.ssrc], stream_state)
+      {actions, state}
+    else
+      stream_state = %State.StreamState{
```
varsill, line 363: [NIT] It would also be fine this way you Static-Fanatic ;)
```suggestion
      stream_state = %{
```

lib/membrane/rtp/demuxer.ex
```diff
@@ -309,55 +321,87 @@ defmodule Membrane.RTP.Demuxer do
...
 
-  defp initialize_new_stream_state(pad_waiting_for_stream, packet, ctx, state) do
+  @spec bind_pad_with_stream(
+          Pad.ref(),
+          output_pad_options(),
+          State.StreamState.t(),
+          RTP.PayloadFormat.payload_type_mapping()
+        ) :: {[Membrane.Element.Action.t()], State.StreamState.t()}
+  defp bind_pad_with_stream(pad, pad_options, stream_state, payload_type_mapping) do
+    if stream_state.pad_match_timer != nil, do: Process.cancel_timer(stream_state.pad_match_timer)
```
varsill, line 385: What are the "canceling" guarantees of that `Process.cancel_timer`? What happens if the `:pad_match_timeout` is already in the message box and it gets cancelled here?

varsill (approved):

test/membrane/rtp/demuxer_muxer_integration_test.exs
Noarkhh: I see your point, but I think that adding constants from the RFC as parts of the API would only clutter it, I don't see anyone calling that function to figure out the size of the timestamp field in a RTP packet

varsill (approved):

PR #183, opened 2026-05-19
Add options to set initial sequence number and timestamp in RTP.Muxer

Noarkhh:
Sometimes it's needed to know what these values will be, for example an RTSP Server needs to respond to a PLAY request with a response containing a `RTP-Info` header containing these values.


lib/membrane/rtp/muxer.ex
```diff
@@ -182,26 +208,29 @@ defmodule Membrane.RTP.Muxer do
...
 
     rtp_offset =
-      buffer.pts
-      |> Membrane.Time.as_seconds()
-      |> Numbers.mult(stream_state.clock_rate)
-      |> Ratio.trunc()
+      div(
+        buffer.pts * stream_state.clock_rate + div(Membrane.Time.second(), 2),
+        Membrane.Time.second()
+      )
```
mat-hek: I liked the previous version more :P

mat-hek (dismissed):

lib/membrane/rtp/muxer.ex
```diff
@@ -37,6 +37,24 @@ defmodule Membrane.RTP.Muxer do
...
+        """
+      ],
+      initial_timestamp: [
+        spec: ExRTP.Packet.uint16() | :random,
+        default: :random,
+        description: """
+        RTP Timestamp that will be assigned to the first packet of this stream. Timestamps of subsequent 
+        packets will be calculated starting from this value. If not provided, a random free value will
+        be assigned.
+        """
```
mat-hek, line 56: Mention that they should be randomly generated to prevent known payload attacks

